PR #5, opened 2018-04-20
Templated TopologyKey

wajihahmed:


helm/opendj/values.yaml
```diff
@@ -77,6 +77,10 @@ djPersistence: true
 # The hard setting means that do not deploy more than one pod in a node which is critical for HA.
 podAntiAffinity: "soft"
 
+# This is the exact value for TopologyKey. The other possible value is "failure-domain.beta.kubernetes.io/zone"
+# which can be used NOT to schedule the pod on a node in the same zone thus allowing for HA across zones
```
dgoldssfo: "NOT to schedule the pod" is a little unclear

Sugg: "... which can be used so that the pod is not scheduled on a node in the same zone. Specifying TopologyKey lets you configure HA across zones." 

or something like that.

wstrange: LGTM

PR #10, opened 2018-04-25
remove maven pom.xml. switch base java / git images from quay.io to the docker hub

wstrange:


wstrange:
This removes the deprecated pom.xml for maven builds
Version bump for IDM / IG
Base images for java, git - now are on the docker hub

PR #11, opened 2018-04-25
Templated the config file and added the actual one to gitignore

wajihahmed:


wstrange:
I think the .gitignore should be at the root level?

PR #15, opened 2018-04-25
Renamed from full-stack to full-platform to align with marketing

wajihahmed:


wstrange:
I think we should change the name again. Just for giggles.

PR #30, opened 2018-05-08
Cloud 9 - cert-manager

lee-baines:
Added TLS certificate objects to AM/IDM/IG
Added prodTLS value
Added certificate cleanup to remove-all script

### DIFF
--- a/bin/remove-all.sh
+++ b/bin/remove-all.sh
@@ -37,3 +37,4 @@ done
 
 
 kubectl delete job --namespace  ${NAMESPACE} --all
+
--- a/helm/README.md
+++ b/helm/README.md
@@ -94,6 +94,13 @@ ingress host name. The format is:
 
 Note that the details of the ingress will depend on the implementation. You may need to modify the ingress definitions. 
  
+# TLS
+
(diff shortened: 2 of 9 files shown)


lee-baines:
sure agreed. Shame they still don't support wildcard certs.  But at least this works without any additional input.

wstrange:
What happens if just leave the TLS secrets as is, and let people delete
them if they wish

lee-baines:
To be honest, I didn't remove the secrets initially. By keeping the TLS secret there, the certificate was only renewed on every deployment rather than a new 1 requested. Cert manager looks for the secret before requesting.

But letsencrypt have a staging area and a production area.  Production has various rate limits so I thought, for dev work, we should use their staging environment(current default behaviour).  Then customers can request production certificates using the prodTLS flag in the values file. But this then meant I needed 2 different secrets but I couldn't keep 2 different names.  So I decided to delete them everytime.

This isn't a problem if I can centralise it with the wildcard cert . I'll check that as thought it wasn't available.

wstrange:
Lee - if we use wildcard certs, I am wondering if we need the tls.yaml file in each product directory.

AFAIK, we can have a single cert request, and then reference the same tls secret in each namespace

PR #42, opened 2018-05-10
Updated idm-cloudsql chart with changes from 6.0.0 openidm chart

robert-blue:


helm/idm-cloudsql/requirements.lock
```diff
@@ -1,11 +1,6 @@
...
-  import-values: null
-  name: git
-  repository: file://../git
-  tags: null
-  version: 0.1.0
-digest: sha256:0f75859e806bd07163b3bbb809d740609dd8bed946d7198c277a092781ca273b
-generated: 2017-07-24T15:20:17.903471153-06:00
+- name: frcommon
+  repository: file://../frcommon
+  version: 6.0.0
```
wstrange: We took out the frcommon chart in master - as folks got tripped up on not updating the deps, and it so trivial .

If there any template expansions you need - just copy them in to your own _helpers.tpl

wstrange: You can delete the requirements.*, and remove any nested charts/*

PR #46, opened 2018-05-11
cloud-453

wstrange:
restore cmp-platform chart


helm/cmp-platform/requirements.lock
```diff
@@ -0,0 +1,24 @@
+dependencies:
+- name: openam
+  repository: file://../openam
+  version: 6.0.0
```
dgoldssfo: Just verifying that 6.0.0 is the correct version here and elsewhere in this file.

wstrange:
The lock file should get regenerated when the helm charts are published.
This should be OK

PR #51, opened 2018-05-15
Dev

wajihahmed:


samples/deployment/type/m-cluster/frconfig.yaml
```diff
@@ -0,0 +1,11 @@
+
+replicaCount: 1
+
+git:
+  #repo: "https://github.com/ForgeRock/forgeops-init.git"
+  repo: git@github.com:ForgeRock/cloud-deployment-config.git
+  branch: master
```
dgoldssfo: Just verifying you actually want to use master and not release/6.0.0

But it looks like we don't have a release/6.0.0 branch. I'm wondering if this might be a problem when the AM configuration starts saving - if so, you want to get Lee to save off a cloud-deployment-config that is known to work with 6.0.0 so that the product doesn't change out from under you.

wajihahmed: Actually it is a directory called R6.0

PR #55, opened 2018-05-16
Update deploy.sh script

wstrange:
PTAL 

bin/deploy.sh samples/config/dev 

Will deploy a dev instance to the "sandbox" namespace

More docs to follow


samples/config/dev/common.yaml
```diff
@@ -0,0 +1,9 @@
+domain: .forgeops.com
+
+image:
+  # repository: gcr.io/engineering-devops
```
dgoldssfo: Might want to switch to images at bintray.io for any public samples so as not to confuse users who find their way here.

PR #58, opened 2018-05-17
Refactor

wajihahmed:


samples/config/prod/deploy.sh
```diff
@@ -84,7 +84,7 @@ deploy_charts()
 {
     echo "=> Deploying charts into namespace \"${NAMESPACE}\""
     # Update openam chart dependencies
-    helm dep up ../../helm/openam
+    helm dep up ../../../helm/openam
```
wstrange: helm dep up should not be required anymore

samples/config/prod/deploy.sh
```diff
@@ -99,8 +99,8 @@ deploy_charts()
             ;;
         esac
 
-        helm install --name ${c}-${NAMESPACE} -f type/common.yaml -f type/${TYPE}/${c}.yaml \
-            --namespace=${NAMESPACE} ../../helm/${chart}
+        helm install --name ${c}-${NAMESPACE} -f common.yaml -f ${TYPE}/${c}.yaml \
+            --namespace=${NAMESPACE} ../../../helm/${chart}
```
wstrange: We should test for the type specific yaml, and if it does not exist (not all will need it) - dont reference it.

samples/config/prod/deploy.sh
```diff
@@ -9,10 +9,10 @@
 
 #### Variables
 
-NAMESPACE="bench"
+NAMESPACE="prod"
 DOMAIN="frk8s.net"
 URL_PREFIX="openam"
-TYPE="m-cluster"
+TYPE="s-cluster"
```
wstrange: We are going to get these from an env.sh file, right? 

At a miniumum we should allow overrides
DOMAIN=${DOMAIN:example.com}

PR #69, opened 2018-05-18
www forgeops landing page

wstrange:
Simple html landing page for forgeops.

Creates a page at http://namespace.domain

### DIFF
--- a/etc/cloud-build-push-charts.sh
+++ b/etc/cloud-build-push-charts.sh
@@ -22,7 +22,7 @@ dir=/tmp/charts
 rm -fr $dir
 mkdir -p $dir
 cd $dir
-charts="frconfig opendj amster openam openidm openig postgres-openidm cmp-platform"
+charts="frconfig opendj amster openam openidm openig postgres-openidm cmp-platform web"
 for chart in $charts
 do
     echo "Packaging $chart"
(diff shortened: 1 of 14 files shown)


etc/cloud-build-push-charts.sh
```diff
@@ -22,7 +22,7 @@ dir=/tmp/charts
 rm -fr $dir
 mkdir -p $dir
 cd $dir
-charts="frconfig opendj amster openam openidm openig postgres-openidm cmp-platform"
+charts="frconfig opendj amster openam openidm openig postgres-openidm cmp-platform web"
```
dgoldssfo: "web" is a pretty generic name - you might want to swap it out for something a little more specific. "forgeops-ui" occurs to me but it's just a thought. No big deal but if this lives "web" might not be the best name long term.

Otherwise, looks great and I'm looking forward to giving it a try.

helm/web/Chart.yaml
```diff
@@ -1,5 +1,6 @@
 apiVersion: v1
 appVersion: "1.0"
-description: A Helm chart for Kubernetes
+description: Simple Landing Page for ForgeOps
```
shankar-forgerock: Is it meant to verify completion of helm install or likely to have more purpose to showcase authn / authz ? [It may be good to indicate the purpose]

wstrange: Right now - just a landing page

shankar-forgerock: Thanks, Would like some visible cues such as landing page rather than kubectl get pods kind of things.

PR #71, opened 2018-05-22
Moving bits from forgeops-internal

Atomic182:
Scope of this PR: 
 - Move benchmark chart to helm/
 - Move gatling image to docker/ 
 - Add smoke tests to cicd/tests/ 
 - Example codefresh yaml with running IG deployment and smoke test run (Do we want to keep that here or should we remove it?)
 - Add simulations to docker/gatling/simulations
 - Modified cloudbuild.yaml and build.sh to contain gatling docker image.


cicd/tests/forgeops-smoke-test.sh
```diff
@@ -0,0 +1,122 @@
...
+POSITIONAL=()
+
+while [[ $# -gt 0 ]]
+do
+key="$1"
+
+case $key in
+    -s|--suite)
+    SUITE="$2"
+    shift
```
wajihahmed: Do you need 2 shifts here?

helm/gatling-benchmark/README.md
```diff
@@ -0,0 +1,68 @@
...
+This helm chart have two purposes. First one is to run actual tests, second one
+to expose results that can be viewed/downloaded
+
+## Setup
+
+### Prerequisites
+This helm chart expects product deployment to be running and configured.
+
+### Gatling benchmark image
+You can either get gatling benchmark image for bintray forgerock repository
```
Atomic182: s/for/from

helm/gatling-benchmark/README.md
```diff
@@ -0,0 +1,68 @@
...
+or build it by yourself. Docker image dockerfile can be found in `forgeops/docker/gatling`.
+To add custom benchmarks, you need to add them to `forgeops/docker/gatling/simulations` folder.
+
+### Benchmark configuration
+Benchmark configuration can be found in values.yaml file.
+You can configure all necessary test related variables here.
```
Atomic182: s/here/there

cicd/tests/testcases/am-smoke.sh
```diff
@@ -0,0 +1,70 @@
...
+          -o /dev/null -w '%{http_code}\n' -s)
+  if [ "$RES" = "200" ]; then
+    PASS=true
+  fi
+}
+
+# Admin login - Login into AM as Amadmin, expect tokenId to be present in resp.
+admin_login_test() {
+  TESTNAME="AM Login test"
+  RES=$( curl --request POST --header \
```
wajihahmed: I usually change the Authentication->Settings->Organization Authentication Configuration to point to my own service.  This means that you can no longer login into AM as "amadmin" using standard json/authenticate endpoint.  You have to use something like json/authenticate/?authIndexType=service&authIndexValue=adminconsoleservice.  So lets make it a habit to call the endpoitn with these query params as a standard practice in our scripts

cicd/tests/testcases/dj-smoke.sh
```diff
@@ -0,0 +1,20 @@
...
+#
+# Each test must set PASS=true/false and optionally RES with content that
+# will be logged in out.txt
+#
+################################################################################
+
+# Testcase definitions
+
+# Package funtions
+tests=(  )
```
wajihahmed: Will this be done later?

docker/gatling/Dockerfile
```diff
@@ -0,0 +1,17 @@
+# Dockerfile for building gatling image with benchmark Tests
+FROM denvazh/gatling:2.2.1
```
wajihahmed: We should consider moving to latest version of gatling as i have been using it without any issues

docker/gatling/custom-gatling.sh
```diff
@@ -0,0 +1,70 @@
+#!/bin/sh
```
wajihahmed: Is there a reason why using "sh" instead of "bash"?

cicd/tests/testcases/dj-smoke.sh
Atomic182: Yes, this is a placeholder for adding a DJ smoke tests.

In current state, I think we can assume basic DJ deployment works if AM/IDM is working.

docker/gatling/Dockerfile
Atomic182: OK, Will swap it to latest. That's 2.3.1 right?

cicd/tests/testcases/am-smoke.sh
Atomic182: In that case we will need to make this change to default AM config as well. I will speak to Lee if it's OK to change it. 

This request is the same as XUI does, right ?

PR #72, opened 2018-05-22
DS - update backup / gcs paths

wstrange:
Scripts update
DS docker image - update backup path

### DIFF
--- a/bin/remove-all.sh
+++ b/bin/remove-all.sh
@@ -32,6 +32,8 @@ pvclist=`kubectl get pvc --namespace ${NAMESPACE} -o jsonpath='{.items[*].metada
 for pvc in ${pvclist}
 do
     echo "Deleting $pvc"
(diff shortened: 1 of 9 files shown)


bin/remove-all.sh
```diff
@@ -32,6 +32,8 @@ pvclist=`kubectl get pvc --namespace ${NAMESPACE} -o jsonpath='{.items[*].metada
 for pvc in ${pvclist}
 do
     echo "Deleting $pvc"
-    kubectl delete pvc ${pvc}
+    kubectl delete pvc --namespace ${NAMESPACE}  ${pvc}
 done
 
+
+kubectl delete job --namespace  ${NAMESPACE} --all
```
shankar-forgerock: Nice.

dgoldssfo: 'Bout time :-)

Thanks, I always for get to do that.

PR #74, opened 2018-05-23
Improvements and fixes for benchmark. 

Atomic182:
- Couple of fixes for benchmark
- Added sample config for benchmark development
- Added missing variables to values.yaml for gatling helm chart
- Fixed wrong gatling docker image build (wrong folder for sims)


docker/gatling/Dockerfile
```diff
@@ -1,17 +1,15 @@
...
 FROM denvazh/gatling:2.3.1
 
-# TODO - Discuss if this is needed now.
-# ADD http://abondance.internal.forgerock.com/pkg/clients/gatling/ext/ldap/gatling-ldap-full-0.0.4-SNAPSHOT.zip /
-# RUN mkdir /ldap-plugin
-# RUN unzip /gatling-ldap-full-0.0.4-SNAPSHOT.zip -d /ldap-plugin
-
-#RUN cp /ldap-plugin/* /opt/gatling/lib/
-# Add local files
+# In case of you want to add custom gatling libraries
```
dgoldssfo: In case you want to add custom Gatling libraries,
make sure you you add them to image and copy to the appropriate place.

docker/gatling/Dockerfile
```diff
@@ -1,17 +1,15 @@
...
+# make sure you you add them to image and copy to apropriate place
+# RUN cp /ldap-plugin/* /opt/gatling/lib/
 
+# Using custom gatling startup scripts for easier passing of variables
```
dgoldssfo: Use custom Gatling startup scripts to make it easier to pass variables through Helm charts.

samples/config/dev-benchmark/README.md
```diff
@@ -0,0 +1,11 @@
+# Dev-benchmark sample config
+This config is used to develop and test gatling benchmark simulation.
```
dgoldssfo: Gatling

In general, FR writing style is to capitalize product names (Gatling, Helm not gatling, helm) when referring to them in text *except* when you're referring to case-sensitive commands (i.e., Run the helm command, not run the Helm command).

docker/gatling/Dockerfile
```diff
@@ -1,17 +1,15 @@
...
+# through helm charts
 COPY custom-gatling.sh /custom-gatling.sh
 RUN chmod +x /custom-gatling.sh
-COPY simulations /opt/gatling/user-files/
+COPY simulations /opt/gatling/user-files/simulations
```
wstrange: Perhaps we should look at loading the simulations via a configmap mounted as a volume?
Helm can slurp up files in a folder and stick them in a configmap

docker/gatling/Dockerfile
Atomic182: Will fix, thanks

docker/gatling/Dockerfile
Atomic182: I agree that will be better for developing and testing benchmarks (no need for rebuilding docker image). I will merge this change and will add this in following set of changes.

samples/config/dev-benchmark/README.md
Atomic182: OK. Thanks for looking on it! Will try to remember this writing style.

PR #75, opened 2018-05-23
Dev

wajihahmed:


samples/config/prod/README.md
```diff
@@ -26,7 +26,7 @@ Other requirements are:
 
 
 ### HA and Multi-Zone deployments
-- If you need HA for a single zone cluster then comment out the *openamReplicaCount* and *replicas* keys in each of the yaml file in the `templates` directory.  Set the value to the desired number but make sure that your node is sized appropiately to support the number.
+- If you need HA for a single zone cluster then comment out the *openamReplicaCount* and *replicas* keys in each of the yaml file in the `templates` directory.  Set the value to the desired number but make sure that your node is sized appropiately to support the number.  Also ensure the value of `ctsStores` and `ctsPassword` is accurate in amster.yaml.
```
shankar-forgerock: It would be good to avoid the term "comment out" as it is confusing [should the comment # flag be set or not?]. I guess in this case the intention is to set the **openamReplicaCount** and **replicas** with appropriate values. So it may be good to write the portion as :  "If you need HA for a single zone cluster then set the desired values for  *openamReplicaCount* and *replicas* keys in each of the yaml file in the `templates` directory.  Make sure that your node is sized appropriately to support the number.

PR #78, opened 2018-05-24
ds updates

wstrange:
Add dstune script, 
update replication purge delay setup to 8 hours
move backup verify option to under backup:

### DIFF
--- a/helm/opendj/values.yaml
+++ b/helm/opendj/values.yaml
@@ -56,6 +56,8 @@ backup:
   # minutes hour day-of-month month day-of-week
   scheduleFull:  "2 2 * * *"
   scheduleIncremental: "45 * * * *"
+  # If true, run a verification job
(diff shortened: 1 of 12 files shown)


helm/opendj/values.yaml
```diff
@@ -56,6 +56,8 @@ backup:
   # minutes hour day-of-month month day-of-week
   scheduleFull:  "2 2 * * *"
   scheduleIncremental: "45 * * * *"
+  # If true, run a verification job
```
shankar-forgerock: It will be great if we output where they can see the result / log of verification job? [especially if they have not or don't want to enable slack notification]

wstrange: If they dont have slack they will have to look at the output of the verify pod log.  The notify.sh script is intended to be replaced if they want with some other notification method

PR #79, opened 2018-05-24
added cert-manager creation script and files

lee-baines:


cert-manager/cluster-issuer.yaml
```diff
@@ -0,0 +1,24 @@
...
+kind: ClusterIssuer
+metadata:
+  name: letsencrypt-prod
+  namespace: kube-system
+spec:
+  acme:
+    # The ACME server URL
+    server: https://acme-v02.api.letsencrypt.org/directory
+    # Email address used for ACME registration
+    email: cloud-deployment@forgerock.com
```
wstrange: just checking - this goes to a real group, right?

cert-manager/cluster-issuer.yaml
```diff
@@ -0,0 +1,24 @@
...
+    # Name of a secret used to store the ACME account private key
+    privateKeySecretRef:
+      name: letsencrypt-prod
```
wstrange: Maybe we should start off with staging, until we know it is really robust

bin/key.json
```diff
@@ -0,0 +1,12 @@
+{
+  "type": "service_account",
```
wstrange: as per discussion, encrypt this with gcloud kms encrypt

bin/deploy-cert-manager.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+
```
wstrange: decrypt key.json.enc - before creating the secret


rm key.json after creating the secret

cert-manager/cluster-issuer.yaml
```diff
@@ -0,0 +1,24 @@
...
+    # Enable the dns01 challenge provider
+    dns01:
+      providers:
+      - name: prod-dns
+        clouddns:
+          project: engineering-devops
+          serviceAccountSecretRef:
+            name: clouddns
```
wstrange: Lets add some README notes or inline comments that briefly explain what is going on here. 
i.e. go to google dns, and do X 
How to create the service account 
what roles the service account needs
etc

cert-manager/cluster-issuer.yaml
lee-baines: well, this did exist but I've tested it and haven't had a response. Have you received anything?

cert-manager/cluster-issuer.yaml
lee-baines: Well, I thought about this, and I couldn't really see any point. We're not paying for anything.  And also, I've tested it over and over again and have also got Pavel to test it and never had an issue with the cert request part itself. I can't see we gain anything by using staging except we have an extra step when loading the AM etc in the browser.  We reuse the same cert so we won't get close to any quotas

bin/key.json
lee-baines: looking at that

PR #81, opened 2018-05-25
Update nfs-provisioner script to kick it, cts index, 

wstrange:
add index for uid=monitor search for the cts
add comment to frconfig that you can use standard storage class on minikube

### DIFF
--- a/helm/opendj/values.yaml
+++ b/helm/opendj/values.yaml
@@ -52,10 +52,9 @@ backup:
   enabled: false
   storageClass: "nfs"
   storageSize: "20Gi"
-  # Backup schedules for full and incrementals. This is a crontab format:
+  # Backup schedule. The first backup is a full, subsequent backups are incremenatal
   # minutes hour day-of-month month day-of-week
-  scheduleFull:  "2 2 * * *"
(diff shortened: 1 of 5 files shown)


helm/opendj/values.yaml
```diff
@@ -52,10 +52,9 @@ backup:
   enabled: false
   storageClass: "nfs"
   storageSize: "20Gi"
-  # Backup schedules for full and incrementals. This is a crontab format:
+  # Backup schedule. The first backup is a full, subsequent backups are incremenatal
```
shankar-forgerock: Add the line to indicate the schedule format is in crontab format.
# The schedule is in crontab format - minutes, hours, day-of-month, month, and day-of-week format

helm/opendj/values.yaml
```diff
@@ -52,10 +52,9 @@ backup:
...
   # minutes hour day-of-month month day-of-week
-  scheduleFull:  "2 2 * * *"
-  scheduleIncremental: "45 * * * *"
+  schedule:  "*/30 * * * *"
```
shankar-forgerock: I think /30 is sufficient. [don't require */30].

PR #91, opened 2018-06-06
Add amster tar.sh to export configuration to a local environment

wstrange:
Set default on ds ha proxy to false

### DIFF
--- a/helm/opendj/values.yaml
+++ b/helm/opendj/values.yaml
@@ -120,11 +120,12 @@ restore:
   # The most recent backup will be restored. 
   enabled: false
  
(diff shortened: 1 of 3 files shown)


helm/opendj/values.yaml
```diff
@@ -120,11 +120,12 @@ restore:
...
   enabled: false
  
+# DS proxy - experimental.
 proxy:
   enabled: false
 
 haproxy:
-  enabled: true
+  enabled: false
   replicas: 1
```
shankar-forgerock: Would we need a replica for proxy ever? if not could we not get rid of haproxy.replicas

PR #93, opened 2018-06-08
New python testing framework for forgeops

Atomic182:
Adding new testing framework and switching smoke tests to use it. 

Additional notable changes: 
- Slack notifications are now send via CURL(easier to change and test), SLACK service url can be configured in trigger(adding custom VAR), so this should be safe not to expose channel url
  ^ We can discuss this
- Added last username that merged changes into master :wink: 

Coming next: 
- HTML report upload & linking results

For notification example, check #cloud-deploy-tests testing room.

### DIFF
--- a/cicd/forgeops-tests/config/ProductConfig.py
+++ b/cicd/forgeops-tests/config/ProductConfig.py
@@ -1,0 +1,67 @@
+"""
+Product config class for tests. Parses env variables into classes and provides them to tests
+Set env variables to override default ones before running tests
+
+Also provides useful generated variables for products (rest endpoints url, etc...)
+
+"""
+
+import os
+
+
+class AMConfig(object):
+    def __init__(self):
+        """
+
+        """
+        try:
+            self.am_url = os.environ['AM_URL']
+        except KeyError:
+            self.am_url = 'http://openam.pavel.forgeops.com/openam'
+
+        try:
+            self.amadmin_pwd = os.environ['AM_ADMIN_PWD']
+        except KeyError:
+            self.amadmin_pwd = 'password'
+
+        self.rest_authn_url = self.am_url + '/json/authenticate'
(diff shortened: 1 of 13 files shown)


cicd/slack-notify.sh
```diff
@@ -0,0 +1,59 @@
...
+fi
+
+if [ -z ${TEST_REPORT_LINK+x} ]; then
+  TEST_REPORT_LINK="No report link"
+fi
+
+
+JSON_SLACK=$(cat <<EOF
+{
+	"mrkdwn": true,
```
Atomic182: indent

cicd/slack-notify.sh
```diff
@@ -0,0 +1,59 @@
...
+  		{
+  			"color": "good",
+  			"text": "Link to report: ${TEST_REPORT_LINK}"
+  		}
+    ]
+}
+EOF
+)
+
+curl --data-urlencode "payload=${JSON_SLACK}" $SLACK_SERVICE
```
Atomic182: ${}

cicd/forgeops-tests/config/ProductConfig.py
```diff
@@ -0,0 +1,67 @@
...
+        self.rest_authn_url = self.am_url + '/json/authenticate'
+        self.rest_oauth2_authz_url = self.am_url + '/oauth2/authorize'
+
+
+class IDMConfig(object):
+    def __init__(self):
+        try:
+            self.idm_url = os.environ['IDM_URL']
+        except KeyError:
+            self.idm_url = 'http://openidm.pavel.forgeops.com/openidm'
```
wstrange: I assume that later we will factor out all the 'pavel' specific paths?

Atomic182: Oh, i will remove them. Good catch. Will go with something generic (openam.example.forgeops.com, etc...)

PR #100, opened 2018-06-19
add new ds dockerfile & chart

wstrange:
add new ds dockerfile & chart
Clean up of files/directories that are no longer used

The old opendj/ image will be retained for a while until we verify the new one works.


docker/ds/Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+
+#####################  Runtime container ################
+FROM openjdk:11-jdk 
+
+COPY --from=0 /var/tmp/bootstrap/run/dsrs1 /opt/opendj/
+
+RUN useradd -d /opt/opendj -G root -m -u 11111 forgerock \
+    && apt-get update && apt-get install -y --no-install-recommends dumb-init vim
+   
+WORKDIR /opt/opendj
```
shankar-forgerock: Should the directory also be moved over to /opt/ds instead of /opt/opendj ?

wstrange: The opendj usage is fairly pervasive throughout the products / scripts. I'm a little hesitant to change it everywhere.

PR #114, opened 2018-06-25
Making default value to install haproxy as false

wajihahmed:


samples/config/prod/m-cluster/userstore.yaml
```diff
@@ -18,5 +18,5 @@ replicas: 2
 topologyKey: "failure-domain.beta.kubernetes.io/zone"
 
 haproxy:
-  enabled: true
+  enabled: false
```
wstrange: should be the new default anyways - but no harm in setting it false

PR #116, opened 2018-06-29
CLOUD-563

wstrange:
use git-init containers


helm/frconfig/values.yaml
```diff
@@ -4,6 +4,13 @@
 
 replicaCount: 1
 
+
+config:
+  name: frconfig
```
dgoldssfo: So the way I read this is that users will still deploy the frconfig chart and it's basically a container for the Git secret and the details of the Git repository. 
Also, when redeploying AM/IDM/IG in the event of a config change, seems like there's no longer the need to reinstall the frconfig chart.
Could you confirm my understanding is correct? Thanks.

helm/amster/templates/amster.yaml
```diff
@@ -20,7 +20,20 @@ spec:
         app: {{ template "fullname" . }}
         component:  {{ .Values.component }}
     spec:
-      terminationGracePeriodSeconds: 5    
+      terminationGracePeriodSeconds: 5   
+      initContainers:
+      - name: git-init
+        image: forgerock/git:6.0.0
```
wajihahmed: should this be 6.5.0?

bin/deploy.sh
```diff
@@ -131,7 +131,7 @@ deploy_charts()
         chart="${comp}"
         case "${comp}" in
           *store)
-            chart="opendj"
+            chart="ds"
```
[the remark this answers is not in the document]
wstrange: You can start to use ds/

helm/frconfig/values.yaml
wstrange: Correct. 
It also means that if they want different config repos (for say IG and IDM), they can deploy another instance of frconfig, with a different config name.

helm/amster/templates/amster.yaml
wstrange: Eventually. Right now there have not been any changes

bin/deploy.sh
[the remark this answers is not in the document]
dgoldssfo: I realize it's going to be a pain, but would you consider changing the names of the openam, openidm, and openig charts so we can have naming consistency? (I'd personally prefer opendj/openam/openidm/openig to ds/openam/openidm/openig)
Not for this PR obviously - I'll create a Jira if you'll consider doing it. Let me know.

PR #135, opened 2018-07-11
Changes to DS build to include IDM repo

alexbugar:


docker/ds/README.md
```diff
@@ -28,15 +28,17 @@ supply the appropriate environment variables in order for the intance to run cor
...
+to use as starter databases.  If data exists in the pvc, the copy will not occur.
 
 The file `env.sh` set a number of environment variables that are used to template out the config.ldif file. For example,
 RS_SERVERS is a list of all the DS/RS hostnames and ports. 
 
 
 ## Limitations
 
-The admin-backed and other LDIF backends currently do not use commons configuration. As a consequence some commands do not work correctly. `dsreplication status`, for example, will not report the correct status.
+The admin-backed and other LDIF backends currently do not use commons configuration. As a consequence some commands do not work correctly. `dsreplication status`,
```
shankar-forgerock: admin-backend ?

docker/ds/bootstrap/setup-ds.sh
```diff
@@ -129,6 +140,12 @@ cp ../../example-v1.json ./config/rest2ldap/endpoints/api
...
 
+
+echo "Putting IDM schema extensions in place"
+cp /var/tmp/schema/* ./db/schema
+
+/var/tmp/bootstrap/setup-idm.sh
+
 ./bin/start-ds
 
 # We only import the ldif on server 1 since we are going to initialize replication from it anywayss.
```
shankar-forgerock: s/anywayss/anyway

shankar-forgerock: s/since we are going to initialize replication from it anywayss/since replication is initiated from server 1/

helm/ds/values.yaml
```diff
@@ -28,14 +29,14 @@ userstore:
...
-  enabled: true
+  enabled: false
 
 image:
   repository: gcr.io/engineering-devops
   # For development we set this to Always to get the most current image.
   pullPolicy: Always
   #pullPolicy: IfNotPresent
-  tag:  6.5.0
+  tag:  alextest
```
dgoldssfo: I assume this is temporary - you might want to add a TODO to change this when the image you want becomes available.

helm/ds/values.yaml
```diff
@@ -14,7 +14,8 @@ useDefaultSecrets: true
 # The setting below changes the base dn for the o=userstore backend only!
 # We strongly suggest that you *NOT* change the default. There is very little value in
 # having the basedn reflect a unique organizational structure.
-baseDN: "o=userstore"
+#baseDN: "o=userstore"
```
dgoldssfo: The kind of contradicts the comment on lines 14-16. If you can't use o=userstore, please change the comment.

docker/openidm2/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM forgerock-docker-public.bintray.io/forgerock/openidm:6.0.0
```
wstrange: I guess openidm2 is for testing? It would be good to have only a single openidm docker image.
A default config is OK here.

Let's put this PR through - but you can do another one that merges the two docker images together.

docker/ds/bootstrap/setup-idm.sh
```diff
@@ -0,0 +1,139 @@
+
```
ludomp: dsconfig has a batch mode.
It would be faster to call dsconfig once with all commands in a batch.

docker/ds/schema/openidm.ldif
```diff
@@ -0,0 +1,115 @@
+#
```
ludomp: From a pure standard point of view, this file contains invalid LDIF.
All attributes of the same types should be next to each other. In other words, put all attributeTypes together, and then all objectClasses.

wstrange: Thanks Ludo!

docker/ds/bootstrap/setup-idm.sh
alexbugar: Good point. Thanks Ludo.

PR #144, opened 2018-07-16
CLOUD-593 Lowered default resources for DS pods to allow for running …

dgoldssfo:
…a smaller clusters

### DIFF
--- a/helm/ds/values.yaml
+++ b/helm/ds/values.yaml
@@ -68,15 +68,17 @@ backup:
 # todo: find JDK 11 args
 #opendjJavaArgs: "-server -XX:+UnlockExperimentalVMOptions -XX:+UseCGroupMemoryLimitForHeap -XX:+UseCompressedOops -XX:+UseG1GC -XX:MaxGCPauseMillis=100 -XX:MaxRAMFraction=2"
 
-opendjJavaArgs: "-Xmx1024m -XX:+UnlockDiagnosticVMOptions"
+opendjJavaArgs: "-Xmx512m -XX:+UnlockDiagnosticVMOptions"
+
 # Resource limits.
-# These help for pod placement on a larger cluster to ensure the OpenDJ instance gets sufficient resources.
-# For production you will want to specify CPU limits to help Kube schedule the pods.
+# These help for pod placement in a larger cluster to ensure the DS instance gets sufficient resources.
+# The default values are artificially low.
+# For production, you will want to increase them.
 resources:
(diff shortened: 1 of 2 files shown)


helm/ds/values.yaml
```diff
@@ -69,14 +69,18 @@ backup:
...
+
 # Resource limits.
-# These help for pod placement on a larger cluster to ensure the OpenDJ instance gets sufficient resources.
-# For production you will want to specify CPU limits to help Kube schedule the pods.
-resources:
- requests:
-  memory: 2000Mi
- limits:
-   memory: 2500Mi
+# These help for pod placement on a larger cluster to ensure the DS instance gets sufficient resources.
```
shankar-forgerock: s/on a larger cluster/in a larger cluster/

helm/ds/values.yaml
```diff
@@ -69,14 +69,18 @@ backup:
...
- requests:
-  memory: 2000Mi
- limits:
-   memory: 2500Mi
+# These help for pod placement in a larger cluster to ensure the DS instance gets sufficient resources.
+# For production, you will want to specify CPU limits to help Kube schedule the pods.
+resources: {}
+
+# Sample values:
+#resources:
```
wstrange: How is this different?

helm/ds/values.yaml
```diff
@@ -69,14 +69,18 @@ backup:
 #opendjJavaArgs: "-server -XX:+UnlockExperimentalVMOptions -XX:+UseCGroupMemoryLimitForHeap -XX:+UseCompressedOops -XX:+UseG1GC -XX:MaxGCPauseMillis=100 -XX:MaxRAMFraction=2"
 
 opendjJavaArgs: "-Xmx1024m -XX:+UnlockDiagnosticVMOptions"
```
wstrange: Also remember to adjust this down as well

helm/ds/values.yaml
dgoldssfo: In the original, we have:

resources:
 requests:
  memory: 2000Mi
 limits:
   memory: 2500Mi

Here I have

resources: {}

(which is what's in the opendj Helm chart)

I was assuming this defaulted to something low but maybe not; do I need to specify an explicit memory amount? (If so, what's the lowest we can go?)

wstrange: we ran into issues with this before during training.  It would be better to set lower limits (and also adjust java opts down)

PR #150, opened 2018-07-20
Cloud 481

lee-baines:
New simplified PR. Have removed original plan to managed TLS certs with a Kubernetes job.  Now there is a simple shell script and template file and a values flag to switch cert manager off.


bin/generate-tls.sh
```diff
@@ -0,0 +1,77 @@
+#!/usr/bin/env bash
+# Script can be used to either automatically generate a self-signed TLS certificate, or to provide your own TLS cert and key.
```
dgoldssfo: Suggestion: This script creates a secret called wildcard.<namespace>.<domain>, which is inherited by the ingress and can be used by all FR components. You can either provide your own TLS cert and key, or you can have the script generate a self-signed TLS certificate.

helm/README.md
```diff
@@ -101,6 +101,13 @@ To enable any of the products to use HTTPS on their external endpoint, set the f
 useTLS: True
 ```
 
+The default behaviour if useTLS = true, is to use certmanager to manage the certificate request/renewal via letsencrypt.  This is enabled by the following flag:
```
dgoldssfo: Sugg: Put this comment above useTLS: true:

The default behavior is to use TLS and have certmanager to manage the certificate request/renewal via 
letsencrypt.

If you want to use TLS but don't want to certmanager to manage the certificate request/renewal, then set
useTLS to true, set useCertManager to false, and run the script ../bin/generate-tls.sh.

helm/openidm/templates/tls.yaml
```diff
@@ -1,20 +1,22 @@
-{{- if .Values.useTLS }}
+# CertManager certificate object created if 'tls.strategy: CertManager'. 
+# This object triggers a certificate request to LetsEncrypt via the CertManager Operator.
```
dgoldssfo: Is it "LetsEncrypt" or "letsencrypt"? Have a look through your comments and pls make this consistent one way or the other.

helm/openidm/values.yaml
```diff
@@ -33,8 +33,10 @@ image:
 # override Java JVM options.
 javaOpts: "-Xmx1024m -server -XX:+UseG1GC"
 
-# If you set this to true, you must have a TLS secret with the same name as the FQDN.
+# TLS is enabled by default. See TLS section for configuration options
```
dgoldssfo: See my suggestion in the README for replacing this comment.

helm/openig/templates/tls.yaml
```diff
@@ -1,20 +1,22 @@
-{{- if .Values.useTLS }}
+# CertManager certificate object created if 'tls.strategy: CertManager'.
```
dgoldssfo: See my suggestion in the README for replacing this comment.

helm/openig/values.yaml
```diff
@@ -28,20 +28,16 @@ config:
 # Tomcat JVM options.
 catalinaOpts: "-Xmx512m"
   
-# If you set this to true, you must have a TLS secret with the same name as the FQDN
+# TLS is enabled by default. See TLS section for configuration options
```
dgoldssfo: See my suggestion in the README for replacing this comment.

bin/generate-tls.sh
lee-baines: Ah yes, this is slightly confusing.  You can actually use the script to upload your own cert and keys too.  You just need to provide the path to the keys.  I'll reword it so its clearer.  So the script can be used in both use cases

lee-baines: I've made it clearer that the script does both

helm/README.md
lee-baines: I changed the 2nd section as suggested but as useTLS isn't enabled by default as part of this PR, I can't change the first section quite yet.  I'll be changing this tomorrow once I've coordinated with Pavel as it impacts his tests.

helm/openidm/values.yaml
lee-baines: done

helm/openig/templates/tls.yaml
lee-baines: done

helm/openig/values.yaml
lee-baines: done

PR #160, opened 2018-07-30
Cloud 617

lee-baines:
Added Prometheus deployment to gke-up script. 
Moved forgerock-metrics under helm folder.
Edited paths so deploy works. 
Added some updated info to kube-prometheus to bring it in-line with the most recent Prometheus release.


bin/gke-up.sh
```diff
@@ -40,3 +41,6 @@ done
 
 # Add cert-manager
 ./deploy-cert-manager.sh
+
+# Add Prometheus
+../monitoring/bin/deploy_prometheus.sh
```
dgoldssfo: Maybe move deploy_prometheus from the monitoring directory to forgeops/bin? I'm not sure why it would still need to be in a separate place.

monitoring/bin/deploy_prometheus.sh
```diff
@@ -1,5 +1,9 @@
 #!/usr/bin/env bash
 
+# Please run from within the Monitoring folder.
+
+MONPATH="../monitoring"
+
```
dgoldssfo: If you move the deploy_prometheus script, this would need to change.

bin/create-cluster.sh
```diff
@@ -14,8 +14,9 @@ echo "Primary Zone = $GKE_PRIMARY_ZONE"
 echo "Additional Zones = $GKE_NODE_LOCATIONS"
 echo "Cluster Name = $GKE_CLUSTER_NAME"
 echo "Cluster Namespace = $GKE_CLUSTER_NS"
+echo "Cluster Monitoring Namespace = $GKE_MONITORING_NS"
```
dgoldssfo: Lee, I think you want to add an entry in ../etc/gke-env.template.cfg for $GKE_MONITORING_NS as well.

bin/gke-up.sh
lee-baines: The only reason I did that was because I was a bit worried about adding another 3 scripts(monitoring/bin/...) into the /bin directory.  It's quite large as it is.  But if you don't think that's an issue then I can do.

bin/create-cluster.sh
lee-baines: the  ../etc/gke-env.template.cfg file is actually a local file and is ignored by helm. That way you don't end up committing your own changes to the repo.  Just copy the template file again

dgoldssfo: Wajih's directions for creating a cluster are to take gke-env.template.cfg and copy it to gke-env.cfg, and make modifications for your environment. So the template file should contain all the variables you need to set when deploying the cluster.

lee-baines: sorry, I was confused because you added cfg at the end.   So yes it should be in the gke-env.template file.  But on the other hand, it works without it as it appears Helm now creates the namespace if it doesn't exist.  I'm sure if never did this.  I'll add it in as a safety net anyway.  It was there originally, not sure what happened.

bin/gke-up.sh
dgoldssfo: Personally, I don't see the point of a monitoring subdir; it would indicate that you deploy the prod sample with monitoring sometimes and without it sometimes. (There's no cert branch for example.)
To have smaller dirs, it might make sense to create some subdirs under bin and etc but I think that's a pretty good sized project.

lee-baines: ok, fair point.  It could easily be removed.  I'll move the values folder under the etc folder.  I'll rename the scripts so they're explicit in their purpose

lee-baines:
That was only because I thought it might be better to keep them together but makes sense. I'll quickly change it now

PR #162, opened 2018-07-30
Making UseTLS true by default and removing deprecated stuff

wajihahmed:
Removing deprecated key-value pairs and fixing paths to repository

### DIFF
--- a/samples/config/prod/m-cluster/ctsstore.yaml
+++ b/samples/config/prod/m-cluster/ctsstore.yaml
@@ -20,7 +20,7 @@ opendjJavaArgs: >
 podAntiAffinity: "hard"
 
 gcs:
-  backup: false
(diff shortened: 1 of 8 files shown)


samples/config/prod/m-cluster/ctsstore.yaml
```diff
@@ -20,7 +20,7 @@ opendjJavaArgs: >
 podAntiAffinity: "hard"
 
 gcs:
-  backup: false
+  backup: true
```
shankar-forgerock: Nice. Thanks.

dgoldssfo: What about creating the NFS backup volumes? Does this get done automatically? Any changes needed to gke-up.sh for this?

PR #163, opened 2018-07-30
added Tomcat filter for CORS

lapinek:
Added Tomcat filter in RS to support CORS requests otherwise rejected. Authentication via cookies is disabled; all requests require bearer token.

### DIFF
--- a/samples/fr-platform/rs/web.xml
+++ b/samples/fr-platform/rs/web.xml
@@ -1,0 +1,60 @@
+<?xml version="1.0" encoding="ISO-8859-1"?>
+<!--
+  The contents of this file are subject to the terms of the Common Development and
+  Distribution License (the License). You may not use this file except in compliance with the
+  License.
+
+  You can obtain a copy of the License at legal/CDDLv1.0.txt. See the License for the
+  specific language governing permission and limitations under the License.
+
+  When distributing Covered Software, include this CDDL Header Notice in each file and include
+  the License file at legal/CDDLv1.0.txt. If applicable, add the following below the CDDL
+  Header, with the fields enclosed by brackets [] replaced by your own identifying
+  information: "Portions Copyright [year] [name of copyright owner]".
+
+  Copyright 2010-2011 ApexIdentity Inc.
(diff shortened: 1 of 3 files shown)


samples/fr-platform/rs/web.xml
```diff
@@ -0,0 +1,4732 @@
...
+        <url-pattern>/cgi-bin/*</url-pattern>
+    </servlet-mapping>
+-->
+
+    <filter>
+    <filter-name>CORSFilter</filter-name>
+    <filter-class>org.apache.catalina.filters.CorsFilter</filter-class>
+    <init-param>
+    <param-name>cors.allowed.origins</param-name>
+    <param-value>*</param-value>
```
dgoldssfo: I thought we didn't use this because it's a security risk

In an example I worked on last year, I have:

http://openam.DOMAIN,http://openidm.DOMAIN where I swapped out DOMAIN for my domain name

samples/fr-platform/rs/web.xml
```diff
@@ -0,0 +1,4732 @@
...
+    <param-name>cors.allowed.headers</param-name>
+    <param-value>Content-Type,X-OpenIDM-OAuth-Login,X-OpenIDM-DataStoreToken,X-Requested-With,Cache-Control,Accept-Language,accept,Origin,Access-Control-Request-Method,Access-Control-Request-Headers,X-OpenAM-Username,X-OpenAM-Password,iPlanetDirectoryPro,authorization</param-value>
+    </init-param>
+    <init-param>
+    <param-name>cors.exposed.headers</param-name>
+    <param-value>Access-Control-Allow-Origin,Access-Control-Allow-Credentials,Set-Cookie</param-value>
+    </init-param>
+    <init-param>
+    <param-name>cors.support.credentials</param-name>
+    <param-value>false</param-value>
```
dgoldssfo: Pls verify this is correct, I have true in my example (but maybe that's wrong)

jakefeasel: The use-case is different here. In this case, we never expect a request to have cookies included. Instead, we expect every request to include an Authorization: Bearer token.

PR #175, opened 2018-08-01
Fix status checks so that they work when a cluster is down

dgoldssfo:
The sanity tests currently in deploy-cert-manager (and some other places in forgeops/bin) do not work when a cluster is temporarily unavailable. They produce messages like this:

The connection to the server 35.231.77.190 was refused - did you specify the right host or port?
./deploy-cert-manager.sh: line 16: [: ==: unary operator expected
tiller is not available...

The fugly "unary operator expected" message happens when you try to compare an empty string to something, which is what we are doing in the script the way it's currently written.

This PR fixes the problem. I'm not the best scripter, therefore it's It's not the most elegant piece of code so if you have suggestions for improving it please make them in comments and I'll update the script.

Wajih, you might want to add code like this (or the improved version) to other scripts.

A better way of doing this might be to have a script that generically tests whether a cluster is available and whether a pod within it is running/available.

Thanks,

David

### DIFF
--- a/bin/deploy-cert-manager.sh
+++ b/bin/deploy-cert-manager.sh
@@ -9,27 +9,47 @@
 # Create secret so the Cluster Issuer can gain access to CloudDNS
 kubectl create secret generic clouddns --from-file=../etc/cert-manager.json -n kube-system
 
-# Need as sometimes tiller is not ready immediately
-while :
+# Check that tiller is running
+while true;
 do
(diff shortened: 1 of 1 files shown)


wajihahmed:
For now this works but the "helm list" is a more definitive test for a working tiller than a running tiller pod.  We might have to add that test as well later.

PR #176, opened 2018-08-01
Remove second confirmation when deleting a cluster

dgoldssfo:
One-liner that removes the requirement of having to type "y" a second time (30-60 after typing the first y) when deleting a cluster.


wajihahmed:
The Y was to acknowledge the warning and give the operator the choice to exit.  Now it is just a warning and if the operator has to exist he/she will have to say No to the next step or Cntrl-C out of it.

PR #177, opened 2018-08-02
transition to nfs client provisioner - WIP

wstrange:
This is a WIP - but can be tested on the shared cluster.

This creates a nfs client provisioner which uses an existing NFS server (cloud filestore). It should be much more reliable than the nfs server provisioner. 

For ds, change the backup.storageClass: nfs-client  
Instead of "nfs"

To try out the new provisioner

### DIFF
--- a/bin/create-cluster.sh
+++ b/bin/create-cluster.sh
@@ -8,6 +8,10 @@
 
 . ../etc/gke-env.cfg
 
+GKE_NETWORK_NAME="${GKE_NETWORK_NAME:-default}"
+GKE_MONITORING_NS="${GKE_MONITORING_NS:-monitoring}"
+
+
 echo "=> Read the following env variables from config file"
 echo "Project Name = $GKE_PROJECT_NAME"
 echo "Primary Zone = $GKE_PRIMARY_ZONE"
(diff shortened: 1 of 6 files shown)


bin/create-cluster.sh
```diff
@@ -8,6 +8,10 @@
 
 . ../etc/gke-env.cfg
 
+GKE_NETWORK_NAME="${GKE_NETWORK_NAME:-default}"
```
dgoldssfo: Hey, I noticed the default network on engineering-docs is Legacy. Should I have a network configuration like that on engineering-docs?

bin/create-cluster.sh
```diff
@@ -70,11 +75,14 @@ gcloud container clusters create $GKE_CLUSTER_NAME \
       --enable-autoscaling \
       --enable-autoupgrade \
       --enable-autorepair \
-      --scopes "gke-default" \
-      --enable-cloud-logging --enable-cloud-monitoring \
+      --scopes  "https://www.googleapis.com/auth/cloud-platform" \
+     --enable-cloud-logging --enable-cloud-monitoring \
```
dgoldssfo: The comment says they're not required, so either the comment is incorrect or you can remove this line.

PR #180, opened 2018-08-02
Perms and dupes

beaucfr:


beaucfr:
We have a configuration data problem we are starting to see as part of saas, enviro variables only get your so far...maybe we can put our heads together and come up with a good solution for all...

PR #181, opened 2018-08-02
update ds to 6.0.0-M1

wstrange:


docker/ds/Dockerfile
```diff
@@ -1,9 +1,9 @@
 
 FROM forgerock/downloader 
 
-ARG VERSION="6.5.0"
+ARG VERSION="6.0.0-M1"
```
dgoldssfo: Just wondering, isn't 6.0.0-M1 the first milestone build for 6.0.0? 

Seems weird

wstrange: oh crap..

PR #190, opened 2018-08-07
Idm init benchmark

Atomic182:
- Added CRUD tests for IDM
- Modified helm chart with gatling to support IDM


docker/gatling/simulations/idm/IDMReadManagedUsers.scala
```diff
@@ -0,0 +1,56 @@
...
+            "X-OpenIDM-Username" -> username,
+            "X-OpenIDM-Password" -> password)
+    }
+
+    val httpProtocol: HttpProtocolBuilder = http
+        .baseURLs(idmUrl)
+        .inferHtmlResources()
+        .contentTypeHeader("""application/json""")
+
+    val readScenario: ScenarioBuilder = scenario("Delete managed users").
```
Atomic182: Delete managed user ->  Read Managed User

Atomic182:
As mentioned by Gary, we need more representative dataset for future benchmarks. We can still get these init tests in, but we will need to improve them in future. Possible way is to use generated users.csv file and mount it for gatling with helm chart(file can be big for 1M+ users)

PR #201, opened 2018-08-09
Some fixs for NFS - CLOUD-640. Omit sidecars for AM by default

wstrange:


helm/dsadmin/values.yaml
```diff
@@ -37,10 +37,9 @@ backup:
   
 # For creating the PV, we need to speficy the server and the path to the export
```
dgoldssfo: s/speficy/specify/

helm/dsadmin/values.yaml
```diff
@@ -37,10 +37,9 @@ backup:
...
 nfs:
-  server: 10.149.108.186
+  server: "10.191.193.66"
```
dgoldssfo: Maybe a comment above this telling users to change this to have the IP address of their own NFS storage

PR #202, opened 2018-08-09
Dev

wajihahmed:


bin/deploy.sh
```diff
@@ -56,6 +56,13 @@ chk_config()
     fi
     echo "=> k8s Context is: \"${context}\""
 
+    if [ "${context}" = "minikube" ]; then
+        echo "=> Minikube deployment detected.  Installing tiller..."
+        helm init --service-account default --upgrade
```
wstrange: I'm not crazy about this being in the deploy script

It should be part of the minikube bringup

bin/deploy.sh
```diff
@@ -158,7 +165,7 @@ deploy_charts()
 isalive_check()
 {
     PROTO="http"
-    if $(grep -q 'useTLS:\s*true' ${CFGDIR}/common.yaml ${CFGDIR}/openam.yaml); then
+    if $(grep -q '^useTLS:\s*true' ${CFGDIR}/common.yaml ${CFGDIR}/openam.yaml); then
```
wstrange: perhaps we should just bite the bullet and not allow http:// at all
It would simplify deployment.
We just need a way to install self signed for minikube

bin/deploy.sh
```diff
@@ -209,9 +216,14 @@ scale_am()
     echo "=> Scaling OpenAM to two replicas..."
     DEPNAME=$(kubectl get deployment -l app=openam -o name)
     kubectl scale --replicas=2 ${DEPNAME}
-    test $? -ne 0 && echo "Could not scale AM pod.  Please check error and fix manually"
+    test $? -ne 0 && echo "Could not scale AM pod.  Please check error and fix manually" 
+}
 
-    printf "\e[38;5;40m=> Deployment is now ready <=\n"
+deploy_hpa()
```
wstrange: should not happen on minikube, right?

samples/config/prod/m-cluster/hpa.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: autoscaling/v1
+kind: HorizontalPodAutoscaler
+metadata:
+  name: openam-hpa
+  #namespace: prod
+  labels:
+    app: openam
+spec:
+  scaleTargetRef:
+    apiVersion: apps/v1beta1
```
wstrange: heads up I think the apps/v1beta1 may change soon

bin/deploy.sh
dgoldssfo: Seems like it shouldn't be that difficult to generate a self-signed cert and install it

PR #205, opened 2018-08-10
Dev

wajihahmed:


samples/config/prod/m-cluster/userstore.yaml
```diff
@@ -1,6 +1,6 @@
 djInstance: userstore
 
-numberSampleUsers: "1000"
+numberSampleUsers: 1000
```
dgoldssfo: I thought we were done with numberSampleUsers in 6.5. Did it creep back in?

wajihahmed: No.  I need to remove it.

PR #209, opened 2018-08-13
Various fixes - dsadmin, IG

wstrange:


javedmshah:
I like that tini was moved to sbin!

PR #216, opened 2018-08-15
Deleted nfs-client-provisioner chart . IDM uses ds as the repo

wstrange:
Deleted nfs-client-provisioner chart - as we no longer need it with FIlestore

Updated openidm.values defaults to use ds as the repo


helm/openidm/values.yaml
```diff
@@ -38,12 +38,18 @@ useCertManager: true
...
-    port: 5432
-    user: openidm
-    password: openidm
-    schema: openidm
-    databaseName: openidm
+    # DS external repo
+    host: userstore-0.userstore
+    port: 1389
+    user: "cn=Directory Manager"
+    password: password
```
shankar-forgerock: Is it OK to have the passwords in clear text?

PR #219, opened 2018-08-17
Bug fixes, enhancements and s-cluster resize

wajihahmed:


docker/ds/scripts/restore.sh
```diff
@@ -29,7 +29,7 @@ done
 if [ !  -z "${OFFLINE}" ]; then
     ARGS="--offline"
 else
-    ARGS="--hostname ${FQDN_DS0} --port 4444 --bindDN \"cn=Directory\\ Manager\" -j ${DIR_MANAGER_PW_FILE} --trustAll"
+    ARGS="--hostname localhost --port 4444 --bindDN \"cn=Directory\\ Manager\" -j ${DIR_MANAGER_PW_FILE} --trustAll"
```
wstrange: I think the restore on boot happens offline, so this may not have the effect you think it does.

PR #224, opened 2018-08-20
Bash best practices and restore fixes and misc

wajihahmed:


bin/deploy.sh
```diff
@@ -180,20 +189,21 @@ isalive_check()
     echo "=> OpenAM is alive"
 }
 
-livecheck_stage1()
+import_check()
 {
-    # This livecheck waits for OpenAM config to be imported.
-    # We are looking to amster pod logs periodically.
-    echo "=> Livecheck stage1 - waiting for config to be imported to OpenAM";
+    # This live check waits for OpenAM config to be imported.
```
dgoldssfo: Not a must fix for now, but in general, could you use "AM" instead of "OpenAM" except when referring to the openam pod? Please change globally.

References to our product names in all UIs, docs, marketing materials etc. have moved away from the names starting with "Open", would be best if we moved away from them in comments. (It would also be best if we changed the name of the openam, openig, openidm pods to am, ig, and idm but there's some overhead associated with that that Warren is not willing to incur.)

bin/deploy.sh
```diff
@@ -243,21 +255,18 @@ create_namespace
 deploy_charts
 
 if [[ " ${COMPONENTS[@]} " =~ " openam " ]]; then
-    echo "AM is present in deployment, running AM livechecks"
-    livecheck_stage1
+    echo "AM is present in deployment, running AM live checks"
```
dgoldssfo: Like this :-)

docker/ds/Dockerfile
```diff
@@ -6,8 +6,8 @@ ARG VERSION="6.5.0-M127.1"
 RUN download -v $VERSION opendj
 
 # JDK 11 tracking - https://bugster.forgerock.org/jira/browse/OPENDJ-5207
-FROM openjdk:11-jdk 
-#FROM openjdk:8-jdk 
+#FROM openjdk:11-jdk 
+FROM openjdk:8-jdk
```
dgoldssfo: It would be really great to stop changing this in our master bits. Maybe folks could create a branch when they want to use JDK 11 in testing?

samples/config/prod/m-cluster/ctsstore.yaml
```diff
@@ -4,22 +4,22 @@ djInstance: ctsstore
...
 
 storageSize: 500Gi
 storageClass: fast
 
 # For JDK 8
 # opendjJavaArgs: >
-#   -server -Xms8g -Xmx8g -XX:+UseCompressedOops -XX:+UseG1GC -XX:+UseNUMA 
+#   -server -Xms10g -Xmx10g -XX:+UseCompressedOops -XX:+UseG1GC -XX:+UseNUMA 
 #   -XX:MaxGCPauseMillis=100 -verbose:gc -XX:+PrintGCDateStamps -XX:+PrintGCTimeStamps 
 #   -XX:+PrintGCDetails -XX:+PrintPromotionFailure -XX:+PrintAdaptiveSizePolicy -Xloggc:/tmp/gc.log
```
dgoldssfo: Just verifying you wanted JDK 11 as default here.

samples/config/prod/s-cluster/env.sh
```diff
@@ -30,4 +30,4 @@ DOMAIN="${DOMAIN/\./}"
 # The components to deploy
 # Note the opendj stores are aliased as configstore, 
 # userstore, ctstore - but they all use the opendj chart
-COMPONENTS=(frconfig dsadmin configstore userstore ctsstore openam amster)
\ No newline at end of file
+COMPONENTS=(frconfig dsadmin configstore userstore ctsstore openam amster openidm)
```
dgoldssfo: :-)

wstrange:
Looks good

lee-baines:
Looks good :)

PR #236, opened 2018-08-27
fr-platform sample README: adding rs-service.sample.svc.cluster.local…

lapinek:
… to local hosts file record resolving to the ingress IP: so that clients running on the localhost could request from the resource server


lapinek:
amended commit: fixed typo

PR #238, opened 2018-08-27
Parameterize additional credentials in IDM and Amster Helm Charts

travishaagen:
Parameterize Prometheus credentials in IDM and Amster Helm Charts, and add optional static user credentials for IDM


travishaagen:
[org-engine: no default passwords for AM/IDM/DS](https://trello.com/c/ozV519mp/594-org-engine-no-default-passwords-for-am-idm-ds)

PR #239, opened 2018-08-27
update monitor  user password from k8s secret

wstrange:


docker/ds/docker-entrypoint.sh
```diff
@@ -34,6 +34,18 @@ update_ds_password()
...
+    if [ ! -f "$MONITOR_PW_FILE" ]; then
+        echo "Can't find the monitor user password file. Won't change the password"
+        return
+    fi
+
+    echo "Updating the monitor user password"
+    pw=`bin/encode-password  -s PBKDF2 -f $MONITOR_PW_FILE | sed -e 's/Encoded Password:  "//' -e 's/"//g' 2>/dev/null`
+    pw="userPassword: $pw"
+    head -n -2  db/monitorUser/monitorUser.ldif >/tmp/pw
+    echo "$pw" >>/tmp/pw
```
dgoldssfo: You sure you want to use append (>>) here rather than overwrite (>)?

PR #244, opened 2018-08-30
postgres and version bumps

wstrange:
Updated postgres helm chart to more current 10.5 release
Updated postgres chart for improved storage class handling
Update AM docker image to M6
Update OpenIDM image to use a snapshot until a new Milestone is produced

### DIFF
--- a/helm/postgres-openidm/values.yaml
+++ b/helm/postgres-openidm/values.yaml
@@ -1,8 +1,4 @@
 
-## postgres image version
-## ref: https://hub.docker.com/r/library/postgres/tags/
-##
-imageTag: "9.5.4"
 
 ## Specify a imagePullPolicy.
 ## 'Always' if imageTag is 'latest', else set to 'IfNotPresent'.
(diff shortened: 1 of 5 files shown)


helm/postgres-openidm/values.yaml
```diff
@@ -23,17 +19,19 @@ imageTag: "9.5.4"
...
 
-## Persist data to a persitent volume.
+## Persist data
 persistence:
   enabled: true
-  storageClass: generic
   accessMode: ReadWriteOnce
-  size: 8Gi
+
+storageClassName: standard
```
wajihahmed: Is this supposed to be at top level?

wstrange: yes -I  moved to top level. There really isn't a good reason to nest this

wajihahmed: Ok but there is the persistence section and this would be a logic fit under it as before?

wstrange: I will top level those as well...

wstrange: done

PR #246, opened 2018-08-31
Enhanced the openidm values.yaml and added appropriate comemtns

wajihahmed:


helm/openidm/values.yaml
```diff
@@ -59,11 +56,12 @@ openidm:
     username: prometheus
     password: prometheus
 
-# Optional - if there is a DJ userstore configured
+# Optional - if there is a DJ userstore configured for synchornization or explict mapping when DJ is used as repo
```
dgoldssfo: s/synchornization/synchronization/
s/explict/explicit/

PR #249, opened 2018-09-01
Update helm charts

alexbugar:
* Update IDM helm chart files to include the IDM-AM integration sample (IDM and AM share the same DS and AM provides authentication for IDM for both ui-admin and ui-user).
* Update AM helm chart files. Added CORS related annotations.
* Update frconfig and amster/am/idm charts to allow global env variables propagation to amster/am/idm containers.

### DIFF
--- a/docker/openidm/Dockerfile
+++ b/docker/openidm/Dockerfile
@@ -1,6 +1,6 @@
 #
 # Copyright (c) 2016-2017 ForgeRock AS.
-FROM forgerock/downloader 
+FROM forgerock/downloader
 
 ARG VERSION="6.5.0-M2"
 RUN download -v $VERSION openidm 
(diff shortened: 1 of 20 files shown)


docker/openidm/Dockerfile
```diff
@@ -1,12 +1,12 @@
...
 # Copyright (c) 2016-2017 ForgeRock AS.
-FROM forgerock/downloader 
+FROM forgerock/downloader
 
 # ARG VERSION="6.5.0-M2"
 # RUN download -v $VERSION openidm 
 
 ARG VERSION="6.5.0"
-RUN download -s -v $VERSION openidm 
+RUN download -s -v $VERSION openidm
```
wstrange: The -s is no longer needed. If you rebase this should get cleaned up

helm/frconfig/templates/config-map.yaml
```diff
@@ -7,4 +7,8 @@ metadata:
 data:
   GIT_REPO: "{{ .Values.git.repo }}"
   GIT_AUTOSAVE_BRANCH:  autosave-{{ .Release.Namespace }}
-  GIT_CHECKOUT_BRANCH:  {{ .Values.git.branch }}
\ No newline at end of file
+  GIT_CHECKOUT_BRANCH:  {{ .Values.git.branch }}
+
+  NAMESPACE: {{ .Release.Namespace }}
```
wstrange: Some charts get the NAMESPACE from the downward API - which is a safer option vs. letting helm set it.

I'd remove NAMESPACE here and set via the api.

PR #253, opened 2018-09-05
Adding "args" to the postgres section 

wajihahmed:
Any arguments to the postgres daemon can be added to this array.  Right now it is exposing two args which can be used to tune the shared_buffers and max_connections.  But any others can be added to this array.  Also the values of the two args is intentionally left at default as they are highly depending on the resources available and in a minimal minikube install we don't want to push the memory.

### DIFF
--- a/bin/create-cluster.sh
+++ b/bin/create-cluster.sh
@@ -72,7 +72,6 @@ gcloud container clusters create $GKE_CLUSTER_NAME \
       --labels="owner=sre" \
       --addons=HorizontalPodAutoscaling \
       --enable-autoscaling \
-      --enable-autoupgrade \
       --enable-autorepair \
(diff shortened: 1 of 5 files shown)


wajihahmed:
Note that in my load testing on m-cluster i am using 256MB shared buffers but did not notice any performance advantage.  This is most likely due to the small data set (about 20k) users that i created during the test using our gatling simulation.  This value *will* matter more in a longer duration test which will cause the postgres database to grow.  Interestingly this buffer cache is analogous to DJ dbcache i.e. it has an overlap with the file system cache.  So with a system with memory constraints such as minikube it is best to leave it at its minimal recommended value of 128MB.  But in a much larger environment it should be tuned according to the work load and anticipated size and growth of the postgres database.

PR #254, opened 2018-09-05
Add cloudbuild scripts to build tag and push artifacts based on a CSV input file

wstrange:
What it says on the tin

Part of CLOUD-711


docker/build.sh
```diff
@@ -102,6 +86,50 @@ if [ -z "$DL" ]; then
   docker build -t forgerock/downloader --build-arg API_KEY=$API_KEY downloader
 fi 
 
+# Take the build list from a CSV file..
```
dgoldssfo: s/.././

docker/build.sh
```diff
@@ -102,6 +86,50 @@ if [ -z "$DL" ]; then
...
+if [ -n "$BUILD_CSV" ]; then  
+  while IFS=, read -r folder artifact tags 
+  do       
+      ${DRYRUN} docker build  $NETWORK --build-arg VERSION=$artifact -t $folder $folder
+
+      # For each registry we support ()
+      for reg in "gcr.io/engineering-devops" "forgerock-docker-internal.bintray.io"
```
dgoldssfo: internal or public?

wstrange: internal is the direction...

PR #271, opened 2018-09-14
Update README.md

c-mccutcheon:
Add --upgrade to helm init command to cater for situations where helm already installed and out of sync. Fixed typo on minikube addon command.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,9 +58,9 @@ the [helm documentation](https://github.com/kubernetes/helm/blob/master/docs/ind
 ```sh
 
 # Make sure you have the ingress controller add on
-minikube addon enable ingress
+minikube addons enable ingress
 
-helm init
+helm init --upgrade
 
 cd helm/
 


README.md
```diff
@@ -58,9 +58,9 @@ the [helm documentation](https://github.com/kubernetes/helm/blob/master/docs/ind
 ```sh
 
 # Make sure you have the ingress controller add on
-minikube addon enable ingress
+minikube addons enable ingress
 
-helm init
+helm init --upgrade
```
dgoldssfo: Hmm. Do we want helm init --upgrade --service-account default in there to allow helm deployment on RBAC-enabled minikubes?

As in https://backstage.forgerock.com/docs/platform/6/devops-guide/#devops-implementation-env-helm-minikube

Clinton, what does the --upgrade do? Should I add it to he official doc?

c-mccutcheon: Started installing locally today in prep for a production deployment, and my client-server tiller versions were out of sync, but installed already. So 'helm init' on its own didn't allow the process to continue unless in the case of a fresh install, helm init --upgrade just enforced the version consistency prior to running the install which then completed successfully.

Good shout on RBAC - I just tested that having --service-account default in the command against a non-RBAC cluster will not allow for charts to be installed without additional cluster role binding setup, I don't think MiniKube has RBAC on by default but could be wrong.

Maybe we should have both options in the readme with a comment? Otherwise need to make the script aware of if RBAC is enabled.

wstrange:
Thanks for the PR!

PR #287, opened 2018-09-23
EKS scripts first cut

wajihahmed:
Additions to forgeops/bin to support Amazon EKS such as creating of cluster, storage class and worker nodes.  Other generic existing scripts will be reused but some older scripts were named too generic so renamed them to be gke specific thus creating space for eks specific ones.  This PR is part of CLOUD-761.


bin/eks-create-sc.sh
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env bash
+# Copyright (c) 2016-2017 ForgeRock AS. Use of this source code is subject to the
```
dgoldssfo: Could we have copyright messages for all our scripts?
When changing existing scripts, the end date should be 2018.
New scripts should be Copyright (c) 2018 ForgeRock AS.

wajihahmed: Yes i will add them in the next PR.

PR #289, opened 2018-09-24
Cloud 775

lee-baines:
Removed sim file copy from gatling docker container.
Added git-init container to mount sim files from forgeops-init.
Added flag and condition to set whether to use /ready file or not.
Updated documentation.

### DIFF
--- a/helm/gatling-benchmark/README.md
+++ b/helm/gatling-benchmark/README.md
@@ -20,8 +20,8 @@ name in values.yaml to match your image.
 Benchmark configuration can be found in values.yaml file.
 You can configure all necessary test related variables there.
 
-## Run benchmark
-To run the benchmark, run following
+## Run benchmark automically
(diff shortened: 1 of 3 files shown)


helm/gatling-benchmark/README.md
```diff
@@ -42,6 +42,13 @@ Benchmark pod will be in Init:0/1 state until tests are finished.
 
 To see gatling progress output, run `./logs.sh`.
 
+## Run benchmark manually
+To manually trigger a benchmark after deployment, you need to follow the following 2 steps:
+* Set ```runAfterDeployment: false``` in gatling-benchmark values.yaml under benchmark section.
+* When you want to run the test, add a ready file at the root of the forgeops-benchmark-gatling container e.g. ```kubectl exe <podname> -c forgeops-benchmark-gatling touch /ready```
```
Atomic182: exe -> exec

PR #295, opened 2018-09-25
Cloud 724 - combined current tls settings into a single setting.

lee-baines:
As previous discussed with David, I've remove the useTLS and useCertManager options and replaced them with a single option tlsStrategy.  This has been added to AM,IG and IDM and tested.

helm/readme updated too.

### DIFF
--- a/helm/README.md
+++ b/helm/README.md
@@ -96,18 +96,17 @@ Note that the details of the ingress will depend on the implementation. You may 
  
 # TLS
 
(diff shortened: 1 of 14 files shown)


lee-baines:
Thanks David, yes I'll change Wajihs deployments too

wstrange:
Let's hold off on this until we see if we can simplify TLS as per our discussion

PR #307, opened 2018-10-01
added config section to galing values file

lee-baines:


helm/gatling-benchmark/values.yaml
```diff
@@ -9,6 +9,14 @@ domain: .forgeops.com
 image:
   repository: forgerock-docker-public.bintray.io/gatling:6.5.0
   pullPolicy: Always
+  
+config:
+  # Name of the configMap that holds the configuration repository URL and of
+  # the secret required to access it.
+  name: frconfig
+  # strategy defines how products get their configuration .
+  # Using the git strategy, each helm chart pulls the configurtion from git using an init container.
```
gandru: typo (2 times) : configurtion -> configuration

lee-baines: I'll correct that in a future PR. Must be like that in all Helm charts as I just copied it

PR #310, opened 2018-10-02
CLOUD-702: Adopt ds 6.5.0 profile creation mechanism to prepare direc…

gary-williams38:
…tory


docker/ds/Dockerfile
```diff
@@ -16,7 +16,7 @@ COPY bootstrap/ /var/tmp/bootstrap/
 COPY schema /var/tmp/schema
 
 # The default BASE_DN of the directory tree.
-ENV BASE_DN o=userstore
+ENV BASE_DN dc=example,dc=com
```
wstrange: do we really want example.com?   I'd prefer a generic "o=userstore" or some such thing.

Ideally organizations don't feel the need to change it

docker/ds/README.md
```diff
@@ -10,7 +10,7 @@ Stage 1 scripts are in the bootstrap/ folder. This folder is not copied into the
 does the following:
 
 * Creates two ds instances that are replicated to each other. The instances have hard coded hostnames (ds1.example.com).
-* Two backends are configured (o=cts, o=userstore) making the instance suitable for a CTS and a userstore
+* Two backends are configured (ou=tokens, dc=example,dc=com) making the instance suitable for a CTS and a userstore
```
wstrange: ditto.

docker/ds/bootstrap/example-v1.json
```diff
@@ -14,7 +14,7 @@
             "subResources": {
                 "users": {
                     "type": "collection",
-                    "dnTemplate": "ou=People,o=userstore",
+                    "dnTemplate": "ou=People,dc=data",
```
gandru: shouldn't it be possible to use @BASE_DN@ ?

gary-williams38: it's on the list of things to try to do later, there are lot of places where suffix i hardwired.

PR #316, opened 2018-10-04
eks-script-fixes

andrewlogan315:
Added route53 record creation to deploy.sh
Fixed set-context using kubectl
Changed permissions on create-ingress-cntlr.sh
Updated warning message on eks-up.sh

### DIFF
--- a/bin/deploy.sh
+++ b/bin/deploy.sh
@@ -178,7 +178,7 @@ isalive_check()
     STATUS_CODE="503"
     until [ "${STATUS_CODE}" = "200" ]; do
         echo "   ${ALIVE_JSP} is not alive, waiting 20 seconds before retry..."
-        sleep 20
+        sleep 10
         STATUS_CODE=$(curl --connect-timeout 5 -k -LI  ${ALIVE_JSP} -o /dev/null -w '%{http_code}\n' -sS || true)
     done
     echo "=> AM is alive"
--- a/bin/deploy.sh
+++ b/bin/deploy.sh
@@ -107,8 +107,8 @@ chk_config()
     echo -e "=>\tComponents: \"${COMPONENTS[*]}\""
 
     AM_URL="${URL_PREFIX:-openam}.${NAMESPACE}.${DOMAIN}"
+ 
 }
-
 
 create_namespace()
 {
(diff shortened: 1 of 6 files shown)


bin/deploy.sh
```diff
@@ -274,6 +285,7 @@ if [ ! -z "$DRYRUN" ]; then
 fi
 
 create_namespace
+create_route53_records
```
dgoldssfo: The intention of deploy.sh is to be cloud-provider agnostic - so I don't think we want to put this call in here

bin/eks-create-cluster.sh
```diff
@@ -55,7 +55,7 @@ do
       echo "EKS cluster is ready"
       aws eks update-kubeconfig --name $EKS_CLUSTER_NAME --kubeconfig ~/.kube/config-eks
       export KUBECONFIG=~/.kube/config-eks
-      kubectx ${CLUSTER_ARN}
+      kubectl config set-context ${CLUSTER_ARN}
```
dgoldssfo: Why? Is kubectx not supported on EKS?

andrewlogan315: We ran into some issues when working with multiple kubernetes environments and this gave us more consistent results

bin/deploy.sh
andrewlogan315: Working on moving this to eks-create-ingress-controller.sh and will update the pull request.

bin/eks-create-cluster.sh
dgoldssfo: Could you please share the problems you hit with the team? Because we are recommending the use of kubectx in docs and using it in our scripts, and if there are problems, perhaps we should move away from it. Thanks.

andrewlogan315: @wajihahmed was the one experiencing most of the inconsistencies with kubectx so he may be able to fill you in better. I had a few instances where I had not cleared my .kube/config-eks after creating a lot of new environments and it would occasionally not switch the context correctly in the script.

wajihahmed: > Why? Is kubectx not supported on EKS?

Kubectx is a third party tool.  It is not part of official kubernetes

dgoldssfo: True, but it is part of CDM - we're telling people to install it and use it, and we're using it in our own scripts.

wajihahmed: Yes but AFAIK it is not part of any scripts in forgeops.  We can certainly mention it as a convenient tool to switch contexts when developing.  However in the scripts, I want to strictly stick to kubectl wherever possible.

dgoldssfo: Ah, you're right. I figured since it was in eks-create-cluster that it had been copied in from gke-create-cluster.
So, would you suggest I pull it out of the docs altogether? It sure is nice for users but if it's flaky on EKS...

PR #322, opened 2018-10-05
Adds PATCH support in RS sample

jakefeasel:
Adds in basic PATCH support for end-user profile maintenance and also a bit of clean-up here and there.


lapinek:
OPTIONS are not allowed for CORS in RS? (I didn't try to run it yet but will do later this weekend.)

PR #324, opened 2018-10-08
add script to bin folder

lee-baines:
Added a simple script which adds commons parameters to exported AM config and also fixes a small bug with site1.json.

### DIFF
--- a/bin/parameterize-am-config.sh
+++ b/bin/parameterize-am-config.sh
@@ -1,0 +1,35 @@
+#!/usr/bin/env bash
+
+# Currently, AM config exported through Amster doesn't retain commons parameters.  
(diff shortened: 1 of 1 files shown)


wstrange:
LGTM

My only question - should config fix up scripts go in the -init repo or here?

dgoldssfo:
> LGTM
> 
> My only question - should config fix up scripts go in the -init repo or here?

I vote for forgeops-init only containing configs. (Just my personal opinion, but IMO "scripts" are "ops" not "init". YMMV.)

lee-baines:
I agree with David for now.  We've moved in a direction where forgeops-init now only contains product configs.  Maybe there's a better place for this but I guess it's all linked.

PR #330, opened 2018-10-09
Enhance debug-logs script to show additional objects. [skip ci] 

wstrange:
CLOUD-821

### DIFF
--- a/bin/debug-logs.sh
+++ b/bin/debug-logs.sh
@@ -2,12 +2,10 @@
 #
 # Simple script to get all logs, descriptions, events from selected namespace.
 # Useful for debugging.
-#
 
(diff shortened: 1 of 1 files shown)


bin/debug-logs.sh
```diff
@@ -2,12 +2,10 @@
...
 # Simple script to get all logs, descriptions, events from selected namespace.
 # Useful for debugging.
-#
 
-SERVICE="service"
-DEPLOYMENT="deployment"
-INGRESS="ingress"
 
+# Max lines to print for pod logs
+NUM_LINES=200
```
dgoldssfo: I'd rather this not be constrained so much (or at all)

PR #336, opened 2018-10-10
Update frconfig to slurp the git ssh secret from a file

wstrange:
Updated the readme with instructions on how to change the secret
Added dummy secrets/id_rsa file

frconfig chart now gets the dummy secret from a file helm/frconfig/secrets/id_rsa.  If you use https git repo, no changes. If you need an ssh secret, either manually replace the created secret, or replace the id_rsa file and deploy


wstrange:
[skip ci]

PR #352, opened 2018-10-16
Updating Skaffold config file using skaffold fix

jascudder:
skaffold.yaml was in old format


jascudder:
Updated after skaffold panic

PR #363, opened 2018-10-22
Updates for smoke tests & deploy.sh

Atomic182:
- Added persistent volume for reports & logs
- Added removal of old(7+ days) reports & logs
- Added IDM live check

### DIFF
--- a/bin/deploy.sh
+++ b/bin/deploy.sh
@@ -106,7 +106,8 @@ chk_config()
     fi
     echo -e "=>\tComponents: \"${COMPONENTS[*]}\""
 
-    AM_URL="${URL_PREFIX:-openam}.${NAMESPACE}.${DOMAIN}" 
+    AM_URL="${URL_PREFIX:-openam}.${NAMESPACE}.${DOMAIN}"
+    IDM_URL="${IDM_URL_PREFIX:-openidm}.${NAMESPACE}.${DOMAIN}"
 }
 
 create_namespace()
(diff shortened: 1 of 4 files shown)


bin/deploy.sh
```diff
@@ -106,7 +106,8 @@ chk_config()
     fi
     echo -e "=>\tComponents: \"${COMPONENTS[*]}\""
 
-    AM_URL="${URL_PREFIX:-openam}.${NAMESPACE}.${DOMAIN}" 
+    AM_URL="${URL_PREFIX:-openam}.${NAMESPACE}.${DOMAIN}"
+    IDM_URL="${IDM_URL_PREFIX:-openidm}.${NAMESPACE}.${DOMAIN}"
```
wstrange: looks good. This might shortly change in the future to a new 'login' prefix - but good for now

bin/deploy.sh
```diff
@@ -183,6 +184,29 @@ isalive_check()
     echo "=> AM is alive"
 }
 
+isalive_check_idm()
+{
+    PROTO="https"
+    if [[ -f ${CFGDIR}/openidm.yaml || -f ${CFGDIR}/common.yaml ]]; then
+        if $(grep -sq -e '^tlsStrategy:\s*\bhttp\b' ${CFGDIR}/common.yaml ${CFGDIR}/openidm.yaml); then
+            PROTO="http"
```
dgoldssfo: I was under the impression that you can no longer use tlsStrategy for http - or did we leave it in just for the smoke tests?

Currently it's an undocumented option.

Atomic182: Right! My bad, I've just copied whatever we had there before. I will remove this and we go with hardcoded https

Atomic182: Removed this for both am & idm

PR #365, opened 2018-10-22
Added EKS scripts for pre-req steps. [skip ci]

djavia:
- Added scripts for documentation
- Fixed typos & cleanup

### DIFF
--- a/etc/amazon-eks-vpc.yaml
+++ b/etc/amazon-eks-vpc.yaml
@@ -1,0 +1,167 @@
+---
+AWSTemplateFormatVersion: '2010-09-09'
+Description: 'Amazon EKS Sample VPC'
+
+Parameters:
+
+  VpcBlock:
+    Type: String
+    Default: 192.168.0.0/16
+    Description: The CIDR range for the VPC. This should be a valid private (RFC 1918) CIDR range.
+
+  Subnet01Block:
+    Type: String
+    Default: 192.168.64.0/18
+    Description: CidrBlock for subnet 01 within the VPC
+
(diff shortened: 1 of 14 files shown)


etc/amazon-eks-vpc.yaml
```diff
@@ -0,0 +1,167 @@
+---
+AWSTemplateFormatVersion: '2010-09-09'
```
shankar-forgerock: Should this be changed to more recent - say '2018-10-23'

etc/amazon-eks-vpc.yaml
```diff
@@ -0,0 +1,167 @@
...
+Description: 'Amazon EKS Sample VPC'
+
+Parameters:
+
+  VpcBlock:
+    Type: String
+    Default: 192.168.0.0/16
```
shankar-forgerock: How does one decide on the IP address and the /16 or /18?

shankar-forgerock: A simple comment on this file covering the following would be great.
Can these parameter values be used unchanged in another EKS environment? Or any values need to be changed?

djavia: Yes that's going to depend on the customer environment. Yes it can be left unchanged and it will create the VPC with the default values.

etc/amazon-eks-vpc.yaml
djavia: No, that should be left as 2010-09-09. 2018-08-30 is the date it was updated/released.

PR #373, opened 2018-10-24
Adding *-up.sh config files for each provider into the *-cluster directory

wajihahmed:


wajihahmed:
After discussing, the "config" directory is not the correct place for these scripts.  Will move them under etc and redo the PR.

PR #374, opened 2018-10-24
Ensure that both legacy and new code works with the mounted keystore secrets

wajihahmed:


helm/openam/templates/config-map.yaml
```diff
@@ -14,8 +14,8 @@ data:
      "dsameUser" : "cn=dsameuser,ou=DSAME Users,{{ .Values.rootSuffix }}",
      "keystores" : {
        "default" : {
-         "keyStorePasswordFile" : "/var/run/secrets/openam/password/.storepass",
-         "keyPasswordFile" : "/var/run/secrets/openam/password/.keypass",
+         "keyStorePasswordFile" : "/var/run/secrets/openam/password/storepass",
```
wstrange: I think this will still cause issues. 

The bootstrap logic will try to rewrite this directory, and will fail with a readonly exception

The files need to be copied to ~/openam/openam

PR #381, opened 2018-10-30
Ds profile update for ou=am-config

wstrange:
WIP - update images for new ds profile for the config store


docker/ds/Dockerfile
```diff
@@ -15,7 +15,7 @@ COPY bootstrap/ /var/tmp/bootstrap/
 COPY schema /var/tmp/schema
 
 # The default BASE_DN of the directory tree.
-ENV BASE_DN dc=data
+ENV BASE_DN dc=ou=identities
```
wajihahmed: Shouldn't this be "ou=identities" without the "dc"

wstrange: good catch. Fixed

PR #399, opened 2018-11-08
Updated tags and repository for all clusters

wajihahmed:


samples/config/prod/l-cluster/ctsstore-local.yaml
```diff
@@ -1,7 +1,6 @@
 image:
-  #repository: forgerock-docker-public.bintray.io/ds
-  repository: gcr.io/engineering-devops/ds
-  tag: 6.5.0-M129.2-CTS-TTL
+  repository: forgerock-docker-public.bintray.io/forgerock/ds
+  tag: 6.5.0-RC4
```
wstrange: Setting the tag will give more stability - so this is fine - but if you want to upgrade in lock step with the helm charts you should omit it.

gary-williams38:
Looks ok, i wonder whether for maintenance we should default to the helm charts for the image tags, repos , etc.

PR #400, opened 2018-11-08
Idm upgrade

jakefeasel:
Updates IDM to 6.5
Aligns the fr-platform sample with the various other changes in forgeops master


samples/fr-platform/README.md
```diff
@@ -69,7 +69,7 @@ Docker, Kubernetes and Helm are used to automate the deployment of this sample.
     grep -v rs-service.sample.svc.cluster.local /etc/hosts \
     | sudo tee /etc/hosts && \
     echo "$(minikube ip) \
-        am-service.sample.svc.cluster.local \
+        login.sample.svc.cluster.local \
```
wstrange: Are we close to being able to make the fqdn a parameter? Tall order- but would be nice

wstrange: Let's say I want to deploy the sample to forgeops.com domain. Is that fairly easy to do, or is the cluster.local hard coded in many places?

PR #405, opened 2018-11-12
Add "||true" to the "rm" command otherwise gke-up fails

wajihahmed:


bin/deploy-cert-manager.sh
```diff
@@ -59,4 +59,4 @@ sleep 5
 kubectl create -f ../etc/cert-manager/cluster-issuer.yaml -n kube-system
 
 # Delete decrypted service account
-rm ../etc/cert-manager/cert-manager.json
\ No newline at end of file
+rm ../etc/cert-manager/cert-manager.json || true
```
wstrange: or rm -f?

PR #414, opened 2018-11-14
Updates fr-platform to use shared helm charts

jakefeasel:
Removes the templates I had been maintaining in favor of direct dependency on the various forgeops provided charts.


alexbugar:
Jake,- in case you want to parameterize global variables "namespace" and "domain" the set up is in place in forgeops/helm/frconfig/templates/config-map.yaml for variable DOMAIN. Variable NAMESPACE is defined as downward API in each product chart.

samples/fr-platform/values.yaml
```diff
@@ -1,11 +1,64 @@
-openam:
+frconfig:
+  domain: .svc.cluster.local
```
jakefeasel: @alexbugar I am specifying the domain value for frconfig here, in my values.yaml. I'm not sure what else you were suggesting I do. Care to elaborate?

alexbugar: The suggestion is about allowing users the flexibility to come up with their own custom values.yaml file where a user can define their own "domain" as a global variable and also use a different "namespace" if they want to.

helm/openidm/templates/idm.yaml
```diff
@@ -59,10 +59,11 @@ spec:
               fieldPath: metadata.namespace
         ports:
         - containerPort: 8080
+        - containerPort: 8444
         livenessProbe:
           httpGet:
             path: /openidm/info/ping
-            port: {{ .Values.service.internalPort }}
+            port: 8080
```
jakefeasel: @wstrange  - is this change okay with you? I made it because I don't want to worry about the k8s client not being able to call the mutual ssl port.

helm/openig/templates/deployment.yaml
```diff
@@ -37,7 +37,7 @@ spec:
       containers:
       - name: {{ .Chart.Name }}
         image:  {{ .Values.image.repository }}:{{ .Values.image.tag }}
-        #imagePullPolicy: {{ .Values.image.pullPolicy }}
+        imagePullPolicy: {{ .Values.image.pullPolicy }}
```
jakefeasel: Heads up - fixed this too

PR #420, opened 2018-11-20
Misc fixes to fr-platform sample

jakefeasel:


samples/fr-platform/amster/config/realms/root/BaseUrlSource.json
```diff
@@ -0,0 +1,19 @@
...
+  "metadata" : {
+    "realm" : "/",
+    "amsterVersion" : "6.5.0-M11",
+    "entityType" : "BaseUrlSource",
+    "entityId" : "BaseUrlSource",
+    "pathParams" : { }
+  },
+  "data" : {
+    "_id" : "",
+    "fixedValue" : "https://login.sample.svc.cluster.local",
```
wstrange: we should figure out how to get rid of this hard coded value. I think commons could be used here?
 &{fqdn}

helm/openidm/templates/openidm-secrets-env.yaml
```diff
@@ -0,0 +1,11 @@
...
+# Common Development and Distribution License (CDDL) that can be found in the LICENSE file
+# Secrets for OpenIDM stack deployment.
+# Note that secret values are base64-encoded.
+apiVersion: v1
+kind: Secret
+metadata:
+   name: "openidm-secrets-env"
+type: Opaque
+data:
+   OPENIDM_ADMIN_PASSWORD: {{ default "openidm-admin" .Values.secret.openidmAdminPassword | b64enc }}
```
wstrange: we should talk to Neil and see how we get rid of the need for passing in this secret value. This feels wrong.

samples/fr-platform/amster/config/realms/root/OAuth2Clients/openidmRS.json
```diff
@@ -24,7 +24,7 @@
       "jwkStoreCacheMissCacheTime" : 60000,
       "idTokenSignedResponseAlg" : "RS256",
       "idTokenPublicEncryptionKey" : null,
-      "jwksUri" : "http://login.sample.svc.cluster.local:80/oauth2/connect/jwk_uri",
+      "jwksUri" : "https://login.sample.svc.cluster.local/oauth2/connect/jwk_uri",
```
wstrange: same deal here. I think commons can be used, and we do make the fqdn available to amster.

samples/fr-platform/idm/conf/authentication.json
```diff
@@ -17,18 +17,30 @@
...
+                    ]
+                },
+                "enabled" : true
+            },
+            {
+                "name" : "STATIC_USER",
+                "properties" : {
+                    "queryOnResource" : "endpoint/static/user",
+                    "username" : "&{ig.client.username}",
+                    "password" : "&{ig.client.password}",
```
wstrange: I gather these are the creds that IG needs to query IDM - hence the need to set the password here.

helm/openidm/templates/openidm-secrets-env.yaml
jakefeasel: Previously this was in a configmap. I figured it was better to have as a secret.

samples/fr-platform/idm/conf/authentication.json
jakefeasel: Yep

lapinek:
Works for me and the skaffold trick works too!

PR #437, opened 2018-11-26
Cloud 370 ig monitoring auth

lee-baines:
added basic auth to servicemonitor for IG.  This needs to be used with the metrics protection filter in IG config as configured in the default IG basic-sample admin.json.

### DIFF
--- a/helm/forgerock-metrics/values.yaml
+++ b/helm/forgerock-metrics/values.yaml
@@ -18,32 +18,32 @@ am:
   enabled: true
   path: /json/metrics/prometheus
   labelSelectorComponent: openam
-  secretUser: cHJvbWV0aGV1cw==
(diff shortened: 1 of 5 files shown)


helm/forgerock-metrics/values.yaml
```diff
@@ -42,8 +42,8 @@ ig:
   enabled: true
   path: /openig/metrics/prometheus
   labelSelectorComponent: openig
-  secretUser: b3BlbmlkbS1hZG1pbg==
-  secretPassword: b3BlbmlkbS1hZG1pbg==
+  secretUser: bWV0cmlj
```
wstrange: are these base64 encoded?
Maybe use the helm b64 function to set the password?

PR #441, opened 2018-11-27
Cloud 913 update amster with commons parameters

lee-baines:
I've added a couple of extra commons parameters to amster configmap as part of updating CDM AM configuration with IG integration.


lee-baines:
there's a couple of extra tweeks in here too for Prometheus and IG

PR #443, opened 2018-11-27
add openig to s-cluster env.sh

lee-baines:
added openig component to env.sh for s-cluster sample. This was missed previously


dgoldssfo:
Not related to this PR, but IG is not currently being scaled like IDM and AM are. This needs to be fixed. I guess the original number of replicas needs to be one and then it needs to be scaled in deploy.sh. Thanks.

PR #456, opened 2018-11-30
(EKS) Hard coding the image.tag for now to avoid load balancing issues…

dgoldssfo:
…s. Sometime in future this shold be taken out once default version is 0.21.0 or greater.


helm/frconfig/secrets/id_rsa
```diff
@@ -1,3 +1,27 @@
-This is a dummy secret for git ssh access. If you are using a private git repo
-that requires an ssh key, replace this file with your generated id_rsa key,
-See the README.
+-----BEGIN RSA PRIVATE KEY-----
```
wstrange: This looks like a real key - we should not be checking in to gi?

dgoldssfo: Sorry 'bout that, thanks for pointing it out. I've removed the file.

PR #459, opened 2018-11-30
Amster userstore configuration

Atomic182:
**Reason:**
People using empty import(empty import is used in docs as well) are currently having issues to create test user because there is wrong suffix passed (ou=am-config). To avoid this, we can define userstore config with amster and pass values from yaml files. 
**Problems:** 
Almost all our current configs(forgeops-init) have id=embedded for userstores and we are not using amster --clean import by default. Using this will now result in having two userstore defined (OpenDJ & embedded)
**Possible solutions to this:**
 - Remove userstore config from all of our configs and define userstore in helm chart
 - Rename existing userstore configs to use id=opendj & filename OpenDJ.json 

I've tested this with empty import as well as existing configs (smoke-tests)

We decided to let this PR on hold until we make 6.5.0 release.

### DIFF
--- a/helm/amster/templates/config-map.yaml
+++ b/helm/amster/templates/config-map.yaml
@@ -12,9 +12,15 @@ data:
     --cfgStore {{ .Values.configStore.type }} \
     {{ if eq .Values.configStore.type "dirServer" -}}
     --cfgStoreHost {{.Values.configStore.host }} \
-    --cfgStoreDirMgrPwd {{.Values.configStore.password }}  \
-    --cfgStorePort {{.Values.configStore.port }}  \
+    --cfgStoreDirMgrPwd {{.Values.configStore.password }} \
+    --cfgStorePort {{.Values.configStore.port }} \
     --cfgStoreRootSuffix {{ .Values.configStore.suffix }} \
+    --userStoreDirMgr "{{ .Values.userStore.dirManager }}" \
+    --userStoreDirMgrPwd  {{ .Values.userStore.password }} \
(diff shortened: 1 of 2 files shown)


dgoldssfo:
Pavel, could we go over the impact of this change? My brain's fried right now.

Thanks,

David

Also, it would help me if you could fill out the new PR template (which didn't exist when you created this PR):

Jira issue? <CLOUD-xxx>
Release 6.5.0 backport required? <yes|no>
6.5.0 doc changes needed? <yes|no>
7.0.0 doc changes needed? <yes|no>
Required README updates made? <yes|no>

Thanks,

David

Atomic182:
Sure David. I'll try to describe the change. If it's unclear, we can sync and i can show what is changed and why(in detail). 

So current state is following: 
By default, when AM is configured without userstore details, configstore is being used as default. This causes problems as configstore uses different suffix for config (ou=am-config) then we are using for users(ou=identities). This works fine when someone imports their own userstore config, but cause problems with empty-import - can't create users due to wrong suffix. 

This change does following: 
- Expose initial userstore config via helm chart
- Defaults are set as before (default userstore is configstore), but setting correct suffix(ou=identities)

Also to fill out template: 
Jira issue - **NONE**
Release 6.5.0 backport required? **Yes**
6.5.0 doc changes needed? **No**
7.0.0 doc changes needed? **No**
Required README updates made? Probably not. I don't see any mentions of configstore configuration in amster readme. I've added comments in same way they are there for configstore settings in amster/values.yaml

PR #467, opened 2018-12-04
added web component to cdm deployments

lee-baines:
Added web component to CDM deployment to allow default IG route to work

### DIFF
--- a/samples/config/prod/l-cluster/env.sh
+++ b/samples/config/prod/l-cluster/env.sh
@@ -30,4 +30,4 @@ DOMAIN="${DOMAIN/\./}"
 # The components to deploy
 # Note the opendj stores are aliased as configstore, 
 # userstore, ctstore - but they all use the opendj chart
-COMPONENTS=(frconfig dsadmin configstore userstore ctsstore openam amster postgres-openidm openidm openig)
+COMPONENTS=(frconfig dsadmin configstore userstore ctsstore openam amster postgres-openidm openidm openig web)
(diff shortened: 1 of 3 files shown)


wajihahmed:
Does this deploy ok? The s-cluster which is very tight on cpu resource.

dgoldssfo:
Does this need to be backported to the release/6.5.0 branch? Does it require any doc changes? Also, it would be great if you could update the Helm README https://github.com/ForgeRock/forgeops/tree/master/helm with a one-liner about the new chart.

lee-baines:
This is not a new Helm chart. I've just included it in the CDM deployment.  I'm working through the CDM cookbook with Gary to see whats been added but this has already been described as 1 of the IG examples in the documentation.  Wajih and I just agreed to not in include the web chart in the deployment, but Gary suggested that it should be in there to capture the default root successfully.

wajihahmed:
No if it is very small then can keep it.  I was just suggesting to try the
full deployment on s-cluster if we have one to be sure.
*--*
*Wajih Ahmed* | * D*istinguished *E*ngineer | *FORGEROCK*
+1.717.379.7777 | http://gotomeet.me/wajih

dgoldssfo:
@lee-baines OK, I see the web helm chart is in 6.5 but it's not in the README and needs to be. Not your problem though.
This change is going to affect the doc, because I will need to add it to the chapter on deploying CDM, and I'll need to change what the users sees when navigating to the IG URL.
I'll need to know if you're planning to backport it to 6.5, in which case I'll need to update the 6.5 docs.

PR #470, opened 2018-12-05
Ig smoke tests imp

Atomic182:
Jira issue? CLOUD-903
6.5.0 doc changes needed? No
7.0.0 doc changes needed? No

Added tests for IG based on m-cluster configuration. Using rs-tokeninfo and rs-tokenintrospect endpoints defined in IG configuration. 

This is integration test between IG and AM

### DIFF
--- a/cicd/forgeops-tests/config/ProductConfig.py
+++ b/cicd/forgeops-tests/config/ProductConfig.py
@@ -76,7 +76,7 @@ class IGConfig(object):
         try:
             self.ig_url = os.environ['IG_URL']
         except KeyError:
-            self.ig_url = 'https://openig.smoke.forgeops.com/'
+            self.ig_url = 'https://openig.smoke.forgeops.com'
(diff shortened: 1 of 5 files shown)


cicd/forgeops-tests/config/ProductConfig.py
```diff
@@ -76,7 +76,7 @@ def __init__(self):
         try:
             self.ig_url = os.environ['IG_URL']
         except KeyError:
-            self.ig_url = 'https://openig.smoke.forgeops.com/'
+            self.ig_url = 'https://openig.smoke.forgeops.com'
```
wstrange: perhaps for another time - but can we make tests so they work in any namespace / domain

wstrange: Ahh - nvm, I see you are getting this from the environment

Atomic182: Yes, these are just defaults.

PR #479, opened 2018-12-07
forgeops: do not create forgerock home in ds Dockerfile as it already…

tomfelsws:
During the ds docker image build, an error message occurs because /home/forgerock already exists at that point, so no need to create it again with useradd -m


tomfelsws:
Sure, this is just a minor cosmetic fix ...
Tom

PR #481, opened 2018-12-07
forgeops: make git and util repos customizable

tomfelsws:
Various helm charts have hardcoded repos for git and util docker images. Added variables to make them customizable, with defaults to the original forgerock repos.


tomfelsws:
Yes, I tested with defining my own values/registry for my test deployment as I had to make a small change in the utils image to make it work in Openshift. I did not test the default values, just copy/pasted them.

Tom

tomfelsws:
Correction: I did make deployments also to GKE with this change using the default values, so all good
Tom

PR #491, opened 2018-12-21
Only run decrypt.sh on GKE as it is specfic to GKE only

wajihahmed:
Jira issue? NONE
Release 6.5.0 backport required? MAYBE
6.5.0 doc changes needed? NO
7.0.0 doc changes needed? NO
Required README updates made? NO

### DIFF
--- a/bin/deploy-cert-manager.sh
+++ b/bin/deploy-cert-manager.sh
@@ -3,8 +3,11 @@
 # Script to deploy Cert-Manager into kube-system namespace.
 # Run ./deploy-cert-manager.sh .
 
-# Decrypt encoded service account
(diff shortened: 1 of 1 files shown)


wajihahmed:
I agree it is not bullet proof.  However for the short term to let this be.  As we add more cloud providers there is too much repetition going on in the scripts.  Therefore I am planning a major refactor of the scripts so that we can have a "common" library of scripts and then much smaller number of provider specific scripts.  Ofcourse we need to discuss all of this in our Sprint planning.

shankar-forgerock:
While on refactoring and rationalizing the scripts I think it would also be
great to locate the provider specific scripts in separate subdirectories
under bin directory. Otherwise there are too many scripts in bin directory.

PR #499, opened 2019-01-04
Backup only specific backends

wajihahmed:
Now backup.sh backs up specific backends instead of backup up all backends.  The reason is that restoring all backends online is not possible which is what scripts/restore.sh does and hence it fails.

Specifically backup.sh now looks at the env vars DS_ENABLE_* and if the value is true adds the backend to the --backendId arg to the DS backup
binary. Hence the backendId are hard coded and if they change in future then this script will have to be modified.

Jira issue? NA
Release 6.5.0 backport required? NO
6.5.0 doc changes needed? NO
7.0.0 doc changes needed? YES
Required README updates made? NO

### DIFF
--- a/docker/ds/scripts/backup.sh
+++ b/docker/ds/scripts/backup.sh
@@ -8,18 +8,35 @@ source /opt/opendj/env.sh
 
 #DATESTAMP=`date "+%Y/%m/%d"`
 
-mkdir -p "$BACKUP_DIRECTORY"
-chmod 775 "$BACKUP_DIRECTORY"
+mkdir -p "${BACKUP_DIRECTORY}"
+chmod 775 "${BACKUP_DIRECTORY}"
+
+IDS=()
+
+if [ "${DS_ENABLE_USERSTORE}" = "true" ]; then
+	IDS+=(" --backendId amIdentityStore")
+fi
+if [ "${DS_ENABLE_IDMREPO}" = "true" ]; then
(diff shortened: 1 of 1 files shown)


shankar-forgerock:
I agree with David.

6.5.0 Doc changes would be required.

-Shankar

wajihahmed:
> https://ea.forgerock.com/docs/platform/gke-cookbook/index.html#cookbook-benchmarking-proc-generate-users

Step 6  above in CDM will *not* require a change because you are restoring a specific backend which if enabled will be backed up by backup.sh

> https://ea.forgerock.com/docs/platform/gke-sr-guide/index.html#sr-backup-user-initiated 

Again within the context of CDM, backup.sh will backup all the backend with data that is changed after the pod comes up such as the amIdentityStore, amCts, cfgRepo and idmRepo.  

The fact is that if you try to run restore.sh on a "full" backup (a backup with all the backends), then it will error out as the "tasks" and "schema" backend need the restore to be done in offline mode which not possible in CDM.

But if we need to create a full backup anyway then the command in the SRE guide will have to change to use either the bin/backup binary or i add a switch to backup.sh to force a full backup.  The third option is to remove the "full backup" verbiage and just call it a back.  Let me know your thoughts...

wajihahmed:
After discussion with David, we will NOT backport to 6.5.0.  He is going to remove any mentions of restore.sh from the SR and CDM guide.

PR #505, opened 2019-01-10
First draft for DS smoke test in which we just ping the instances usi…

csovant:
…ng /alive endpoint

Jira issue? CLOUD-1018
6.5.0 doc changes needed? no
7.0.0 doc changes needed? no

### DIFF
--- a/cicd/forgeops-tests/config/ProductConfig.py
+++ b/cicd/forgeops-tests/config/ProductConfig.py
@@ -5,8 +5,9 @@ Set env variables to override default ones before running tests
 Also provides useful generated variables for products (rest endpoints url, etc...)
 
 """
-
+# Lib imports
 import os
+import subprocess
 
 # Global flag to enable/disable verification of certificates
 try:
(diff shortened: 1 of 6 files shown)


cicd/forgeops-tests/config/ProductConfig.py
```diff
@@ -80,6 +86,39 @@ def __init__(self):
...
+class DSConfig(object):
+    def __init__(self):
+        if is_cluster_mode():
+            self.ds0_url = 'https://userstore-0.userstore:8080'
+            self.ds1_url = 'https://userstore-1.userstore:8080'
+        else:
+            self.helm_cmd = 'kubectl'
+
+            self.ds0_local_port = 8080
+            cmd = self.helm_cmd + ' --namespace csovant port-forward pod/userstore-0 %s:8080' % self.ds0_local_port
```
Atomic182: I see hardcoded namespace here. We should probably make this configurable.

cicd/forgeops-tests/tests/smoke/ds/DSSmoke.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+class DSSmoke(unittest.TestCase):
+    dscfg = DSConfig()
+
+    def test_0_ping(self):
+        """Pings OpenDJ instances to see if servers are alive"""
+        resp = get(verify=self.dscfg.ssl_verify, url=self.dscfg.ds0_rest_ping_url)
+        self.assertEqual(resp.status_code, 200, 'userstore-0 is alive')
+        self.dscfg.dj0_popen.kill()
```
gary-williams38: Killing the port-forward probably not a per test, also only used when not in cluster mode.

csovant:
@markg58 The error you get is because the port forward is not working in your case due to hard-coded value `csovant` for namespace

PR #511, opened 2019-01-13
Idm user cleanup

markg58:
Jira issue? None
6.5.0 doc changes needed? No
7.0.0 doc changes needed? No

Not sure about why commit d93e43b is included with this PR.


cicd/forgeops-tests/tests/smoke/idm/IDMSmoke.py
```diff
@@ -203,3 +204,21 @@ def test_7_user_reset_pw(self):
                         params=params, json=payload4)
         self.assertEqual(200, stage4.status_code, "Stage 4 - Password reset")
         s.close()
+
+    @classmethod
+    def tearDownClass(self):
+        instance = IDMSmoke()
+        """Delete user rsutter"""
```
andrewvinall: User "rsutter" is only created in test `test_6_user_self_registration` and then `test_7_user_reset_pw` is dependent on this user.  So this cleanup would be best done on a test rather than class basis.  It would also be best if tests 6 and 7 are independent of each other

PR #514, opened 2019-01-14
Apply fix for OPENAM-14170 - make sure tomcat reports correct port/sc…

mrpotes:
This should prevent the need for configuring the base url source service, as tomcat should always report the correct URL.

Jira issue? OPENAM-14170
6.5.0 doc changes needed? no
7.0.0 doc changes needed? no

### DIFF
--- a/docker/openam/server.xml
+++ b/docker/openam/server.xml
@@ -44,7 +44,9 @@
                SSLEnabled="false"
                connectionTimeout="20000"
                URIEncoding="UTF-8"
-               redirectPort="8443" />
+	       redirectPort="8443"
+               proxyPort="443"
+               scheme="https" />
 
       <!--  We terminate ssl externally
 


wajihahmed:
Tomcat needs to be http internally.  Wouldn't the "https" scheme break it?  Just want to make sure this change has been deployed and tested.  If yes then please ignore.

mrpotes:
It hasn't been tested, I'm hoping someone can do that for me. The scheme setting is what tomcat uses to construct the external URI to the contained application.

wajihahmed:
There is an ingress controller (nginx reverse proxy) in front  of Tomcat
which is the external URL and SSL is also terminated on the ingress
controller.

*--*
*Wajih Ahmed* | * D*istinguished *E*ngineer | *FORGEROCK*
+1.717.379.7777 | http://gotomeet.me/wajih

mrpotes:
Yes, but the AM application needs to know what URL its request originally came from.

wstrange:
On further testing of this it does not work. 
Tomcat redirects internal requests (http://openam ) to http://openam:443 - which does not have a listener. We would need to set the proxyName to the external FQDN -but this makes the docker image hard to configure for any environment.

PR #518, opened 2019-01-16
CLOUD-1018 Test restructure. Move stop ds port forwarding.

andrewvinall:
Jira issue? CLOUD-1018
6.5.0 doc changes needed? No
7.0.0 doc changes needed? No

This is a tentative PR primarily to open up discussion.  Questions:
* Currently we have 2 tearDownClass methods.  How does this manifest?
* Is it better to separate associated tests into their own classes.  Currently we only have 2 tests, however the user created in the second test has no relevance to the other tests.
* My opinion is that clean up of the DSConfig class should be handled by that class rather than the calling class.  I have found some potential solutions here https://stackoverflow.com/questions/865115/how-do-i-correctly-clean-up-a-python-object and have tried to implement the Python's "with" statement.  However, I seek advice from more experienced Python programmers.

### DIFF
--- a/cicd/forgeops-tests/config/ProductConfig.py
+++ b/cicd/forgeops-tests/config/ProductConfig.py
@@ -104,6 +104,13 @@ class DSConfig(object):
         self.ds1_rest_ping_url = self.ds1_url + '/alive'
         self.ssl_verify = SSL_VERIFY
 
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        self.stop_ds_port_forward(instance_nb=0)
+        self.stop_ds_port_forward(instance_nb=1)
+
     def stop_ds_port_forward(self, instance_nb=0):
         if not is_cluster_mode():
             eval('self.ds%s_popen' % instance_nb).kill()
--- a/cicd/forgeops-tests/tests/smoke/ds/DSSmoke.py
+++ b/cicd/forgeops-tests/tests/smoke/ds/DSSmoke.py
@@ -15,73 +15,68 @@ import logging
 # Framework imports
 from config.ProductConfig import DSConfig
 
-
-class DSSmoke(unittest.TestCase):
-    dscfg = DSConfig()
-
-    @classmethod
-    def tearDownClass(self):
-        """DSSmoke suite teardown"""
-        self.dscfg.stop_ds_port_forward(instance_nb=0)
-        self.dscfg.stop_ds_port_forward(instance_nb=1)
-
+class DSPing(unittest.TestCase):
     def test_0_ping(self):
         """Pings OpenDJ instances to see if servers are alive"""
-        response = get(verify=self.dscfg.ssl_verify, url=self.dscfg.ds0_rest_ping_url)
-        self.assertEqual(response.status_code, 200, 'userstore-0 is alive')
+        with DSConfig() as dscfg:
+            response = get(verify=dscfg.ssl_verify, url=dscfg.ds0_rest_ping_url)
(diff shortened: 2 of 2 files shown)


markg58:
Trying to answer your questions:
- Currently we have 2 tearDownClass methods. How does this manifest?  _There is a tearDownClass() defined in DSSmoke.py and one in IDMSmoke.py.  As each test class can be run separately.  Or am I missing something in your question?_

- Is it better to separate associated tests into their own classes. Currently we only have 2 tests, however the user created in the second test has no relevance to the other tests.  _I don't believe associating the tests within their own classes adds any value.  And just complicates the code more.  Am I the only one with that opinion??_

- My opinion is that clean up of the DSConfig class should be handled by that class rather than the calling class. I have found some potential solutions here https://stackoverflow.com/questions/865115/how-do-i-correctly-clean-up-a-python-object and have tried to implement the Python's "with" statement. However, I seek advice from more experienced Python programmers - _Based upon [class-and-module-fixtures](https://docs.python.org/3/library/unittest.html#class-and-module-fixtures), tearDownClass() is not called on base class unless you call it programmatically.  And DSConfig doesn't have a tearDownClass() either so I would not expect tearDownClass to be called for any class other than DSSmoke._

csovant:
Agree with Mark this complicates the code.

Maybe one thing to change is the second tearDownClass() (the one which delete the user) to move it as a test teardown => to only call it after test_resource_replication testcase but not sure this is possible in unittest

andrewvinall:
Thanks for your feedback guys.  As I said I was hoping to open up discussion.
Primarily my points are:
1. Two tearDownClass methods.
2. Test scenarions should be grouped closely by function and independent of each other.
3. The DSConfig constructor has a side affect of setting up port forwarding.  I don't think it should be the responsibility of the calling class to clean this up.

The job of porting the tests from unittest to pytest may be an opportunity to think a bit about the structure of our tests.


1. We seem to agree that the two tearDownClass methods in a class is an error.

2. So maybe DSSmoke.py was a bad choice for an example; as currently there are only 2 tests.  However, both tests are not functionally closely related to each other.  In future there will be more DS tests added.   The point I was attempting to make is best summarised here https://docs.python.org/3/library/unittest.html#organizing-test-code:
    "The testing code of a TestCase instance should be entirely self contained, such that it can be run either in isolation or in arbitrary combination with any number of other test cases."
    
In DSSmoke.py test_0_ping is not closely related to test_resource_replication.   Also, note running test_0_ping in isolation will fail the tearDownClass.  I agree this could be resolved by using a tearDown method.

IDMSmoke.py may have been a better code example to make my point with.
* test_0_ping bears no relationship to the other tests.  However, test_0_ping passing would be good prerequisite for the other tests.

* There is a strong dependency between may of the tests.  We should make better use of setup and teardown to allow the tests to be independent of each other, so that they can be run in isolation.
    test_0_ping
    test_1_create_managed_user <-- test_2_update_managed_user <-- test_4_login_managed_user
    test_3_run_recon_to_ldap
    test_1_create_managed_user <-- test_5_delete_managed_user
    test_6_user_self_registration <-- test_7_user_reset_pw
    test_6_user_self_registration <--tearDownClass

* tearDownClass will fail if test_6_user_self_registration has not been run.  So test_6_user_self_registration is the only test that can be run in isolation.

3. Port forwarding for DS is setup as a side-affect of the constructor for class DSConfig.  So I don't think it should be the responsibility of the calling class to clean this up.  I was looking for a mechanism to resolve this issue.

andrewvinall:
This work will be dealt with by later stories.

PR #520, opened 2019-01-16
Removes hard-coded platform secrets entry in favor of dynamic option

jakefeasel:
The hard-coded IG_CLIENT_USERNAME and IG_CLIENT_PASSWORD entries are now going to be specified as part of the helm values. In addition to removing this project-specific config from the general-purpose charts, it also allows other secrets to be specified as needed (for example, if someone wants to add a new secret that wasn't predicted, they can do so and it will become available as part of the environment for the various frconfig-dependent pods)


jakefeasel:
Related forgeops-init change: https://github.com/ForgeRock/forgeops-init/pull/151

PR #524, opened 2019-01-17
(CLOUD-953) Migrate test framework to pytest

csovant:
Jira issue? CLOUD-953
6.5.0 doc changes needed? no
7.0.0 doc changes needed? no


cicd/forgeops-tests/tests/smoke/ds/test_ds.py
```diff
@@ -8,32 +8,36 @@
...
     @classmethod
-    def tearDownClass(self):
-        """DSSmoke suite teardown"""
-        self.dscfg.stop_ds_port_forward(instance_nb=0)
-        self.dscfg.stop_ds_port_forward(instance_nb=1)
+    def setup_class(cls):
+        """DSSmoke suite setup"""
+
+        cls.dscfg.ds0_popen = cls.dscfg.start_ds_port_forward(instance_nb=0)
+        cls.dscfg.ds1_popen = cls.dscfg.start_ds_port_forward(instance_nb=1)
```
andrewvinall: Nice to have this here.

cicd/forgeops-tests/tests/smoke/ds/test_ds.py
```diff
@@ -8,32 +8,36 @@
...
+class TestDS(object):
     dscfg = DSConfig()
 
     @classmethod
-    def tearDownClass(self):
-        """DSSmoke suite teardown"""
-        self.dscfg.stop_ds_port_forward(instance_nb=0)
-        self.dscfg.stop_ds_port_forward(instance_nb=1)
+    def setup_class(cls):
+        """DSSmoke suite setup"""
```
andrewvinall: Comment is out of date.
Is the comment (and similar elsewhere) even necessary?

cicd/forgeops-tests/tests/smoke/idm/test_idm.py
```diff
@@ -8,28 +8,28 @@
...
+
+        logger.test_step('Ping OpenIDM')
+        response = get(verify=self.idmcfg.ssl_verify, auth=('openidm-admin', 'openidm-admin'),
+                       url=self.idmcfg.rest_ping_url)
+        rest.check_http_status(http_result=response, expected_status=200)
 
     def test_1_create_managed_user(self):
         """Test to create managed user as admin"""
+
         headers = self.idmcfg.get_admin_headers({
```
andrewvinall: Future work: Would be good to use this get_ccccc_headers consistently for the other products too.

cicd/forgeops-tests/tests/smoke/ds/test_ds.py
```diff
@@ -56,32 +60,39 @@ def test_resource_replication(self):
...
-    @classmethod
-    def tearDownClass(self):
-        instance = DSSmoke()
-        """Delete datastore resource from users/"""
+    @pytest.fixture()
+    def setup_teardown_test_resource_replication(self):
+        """"Setup and Teardown for test_resource_replication"""
 
-        logging.info("Deleting user entry with ID newuser_ds0 from userstore-0")
+        yield "resource"
```
andrewvinall: I haven't grasped fixtures yet.  But does "resource" map to test_1_resource_replication?

cicd/forgeops-tests/tests/smoke/idm/test_idm.py
markg58: I don't think other products pass headers the same way that IDM does.

cicd/forgeops-tests/Dockerfile
```diff
@@ -3,6 +3,10 @@ FROM python:3.5-alpine
 
 RUN apk add --no-cache bash
 
+RUN pip3 install pytest
+RUN pip3 install allure-pytest
+RUN pip3 install pytest-html
+RUN pip3 install pytest-metadata
+RUN pip3 install requests
 RUN pip install requests
```
markg58: remove this line as previous line is ```RUN pip3 install requests```

cicd/forgeops-tests/Dockerfile
```diff
@@ -3,6 +3,10 @@ FROM python:3.5-alpine
...
-RUN pip install html-testRunner
 RUN pip install beautifulsoup4
```
markg58: should this be ```RUN pip3 install beautifulsoup4```

cicd/forgeops-tests/config/ProductConfig.py
```diff
@@ -11,6 +11,9 @@
 import time
```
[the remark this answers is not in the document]
markg58: Are we going to follow the PEP8 style guide?

- [Class Names](https://www.python.org/dev/peps/pep-0008/#id41)
- [Package and Module Names](https://www.python.org/dev/peps/pep-0008/#id40)

cicd/forgeops-tests/config/ProductConfig.py
```diff
@@ -97,8 +100,10 @@ def __init__(self):
             self.ds1_url = 'http://userstore-1.userstore.%s.svc.cluster.local:8080' % tests_namespace()
         else:
             self.helm_cmd = 'kubectl'
-            (self.ds0_url, self.ds0_popen) = self.start_ds_port_forward(instance_nb=0)
-            (self.ds1_url, self.ds1_popen) = self.start_ds_port_forward(instance_nb=1)
+            self.ds0_local_port = self.get_free_port(8080)
+            self.ds0_url = 'http://localhost:%s' % self.ds0_local_port
+            self.ds1_local_port = self.get_free_port(8080)
+            self.ds1_url = 'http://localhost:%s' % self.ds1_local_port
```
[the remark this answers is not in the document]
markg58: Is there an advantage to have the class (DSConfig) attributes in a collection?  I could possibly see that if they were part of ProductConfig or were being shared between classes, but they aren't.  But, I could be mistaken.  Liked to understand more.

cicd/forgeops-tests/tests/smoke/ds/test_ds.py
csovant: All these messages are docstrings and appear in the report (need to do modification to have them in the html one but we have them in allure for sure) as decription for the testcase.
Can be also used by pydoc to generate a doc for the tests.

cicd/forgeops-tests/tests/smoke/ds/test_ds.py
csovant: Yes in fact it seems that you can put what you want in the string. 
I used "resource" because I found that in an example but you can use what you want, I don't where this info/string appear.
Looks like the only important stuff is the `yield` call which calls the calling method.

cicd/forgeops-tests/Dockerfile
csovant: Good catch, should be `pip3` in this file

cicd/forgeops-tests/config/ProductConfig.py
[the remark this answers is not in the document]
andrewvinall: PEP8 style guide seems a good choice.

cicd/forgeops-tests/tests/smoke/idm/test_idm.py
andrewvinall: This appears in test_am.py
`    def setup_class(cls):
        """Setup a user that will be tested in user login."""

        headers = {'X-OpenAM-Username': 'amadmin', 'X-OpenAM-Password': 'password',
                   'Content-Type': 'application/json', 'Accept-API-Version': 'resource=2.0, protocol=1.0'}`

It will be good to try and be consistent.

PR #539, opened 2019-01-29
Add CA cert to keystore so keytool can establish CA trust

simonhardingforgerock:
Sorry about the lack for Jira for this :)

As the comment a the top of this script states, this script doesn't quite work. It fails with:
`keytool error: java.lang.Exception: Failed to establish chain from reply`

This PR fixes the script by adding the CA cert to the keystore, allowing keytool to establish CA trust.

### DIFF
--- a/docker/ds/bootstrap/create-keystores.sh
+++ b/docker/ds/bootstrap/create-keystores.sh
@@ -48,6 +48,17 @@ create_keystores()
             -dname "$SSL_CERT_CN" \
             -keypass:file $KEYSTORE_PIN
 
+   echo "CA_CERT is $CA_CERT"
+
+    keytool -import \
+            -alias ${CA_CERT_ALIAS} \
+            -trustcacerts \
+            -noprompt \
+            -deststorepass:file ${KEYSTORE_PIN} \
+            -destkeystore ${SSL_KEYSTORE} \
+            -file ${CA_CERT} \
+            -srcstorepass:file ${KEYSTORE_PIN}
+
     keytool -keystore $SSL_KEYSTORE \
             -storetype PKCS12 \
             -storepass:file $KEYSTORE_PIN \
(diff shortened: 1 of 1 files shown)


dgoldssfo:
@simonhardingforgerock When you created the PR, did the template come up?

simonhardingforgerock:
> @simonhardingforgerock When you created the PR, did the template come up?

Yes it did. I was just being lazy and not doing the Jira. See #cloud-deployment in slack.

dgoldssfo:
No Jira is fine, just put n/a or something in the field that asks for it. But it's really helpful for me to have the information about whether any doc changes are needed, and it's helpful for the team to know whether the fix should be backported to the release branch.

simonhardingforgerock:
> No Jira is fine, just put n/a or something in the field that asks for it. But it's really helpful for me to have the information about whether any doc changes are needed, and it's helpful for the team to know whether the fix should be backported to the release branch.

Ok no probs David, I'll elaborate:
- This PR fixes the create-keystores.sh script for DS.
- The purpose of the script is to create a CA and SSL certs that are signed by that CA for use with DS. If you want to replace the default secrets in the charts then it really helps to run this script, as the charts don't work without all of the secrets in place, even if you are not using SSL in the back end.
- This script as far as I know is undocumented. In general, I don't think there is any documentation on how to go into production without using the default secrets (although I know there are several people looking at different approaches to take here).
- It was useful to me as a customer of ForgeOPS 6.5 (Open Banking) to use a fixed version of this script initially, so I think it would be worth back porting.

dgoldssfo:
Thanks, Simon!

I don't mean to be a pain, but at this point in forgeops history, the docs team has to read every line of every PR to assess whether a doc change is needed. That's not always as easy to do as you might think. So we instituted the template so that PR submitters can lend a hand when they change the code base.

PR #543, opened 2019-01-30
Updates end-user-ui port to run on 8080 instead of 8888

jakefeasel:
Using 8888 was causing some annoying conflicts with skaffold and the example oauth2 clients. It's just easier to have the end-user-ui use 8080 like most of our other web apps.

### DIFF
--- a/docker/end-user-ui/Dockerfile
+++ b/docker/end-user-ui/Dockerfile
@@ -13,6 +13,6 @@ RUN git checkout 6.5.0 && npm install && npm run build
 
 WORKDIR /tmp/end-user-ui/dist
 
-EXPOSE 8888
+EXPOSE 8080
 
-CMD ["http-server", "-p", "8888"]
+CMD ["http-server", "-p", "8080"]
--- a/helm/end-user-ui/templates/deployment.yaml
+++ b/helm/end-user-ui/templates/deployment.yaml
@@ -25,7 +25,7 @@ spec:
           imagePullPolicy: {{ .Values.image.pullPolicy }}
           ports:
             - name: http
-              containerPort: 8888
+              containerPort: {{ .Values.service.internalPort }}
               protocol: TCP
           livenessProbe:
             httpGet:
--- a/helm/end-user-ui/values.yaml
+++ b/helm/end-user-ui/values.yaml
@@ -19,7 +19,7 @@ service:
   name: end-user-ui
   type: ClusterIP
   externalPort: 80
-  internalPort: 8888
+  internalPort: 8080
 
 
 ingress:


docker/end-user-ui/Dockerfile
```diff
@@ -15,4 +15,4 @@ WORKDIR /tmp/end-user-ui/dist
 
 EXPOSE 8888
 
-CMD ["http-server", "-p", "8888"]
+CMD ["http-server", "-p", "8080"]
```
markcraig: 8080 is fine with me... but what about the IG welcome page currently on localhost:8080?

Here's what I get after the `skaffold dev` step in <https://github.com/ForgeRock/forgeops-init/tree/master/6.5/oauth2/development>:

```
$ lsof -i :8080
COMMAND  PID USER   FD   TYPE DEVICE SIZE/OFF NODE NAME
kubectl 9822 mark    6u  IPv6 125886      0t0  TCP ip6-localhost:http-alt (LISTEN)
kubectl 9822 mark    7u  IPv4 125887      0t0  TCP localhost:http-alt (LISTEN)
$ curl http://localhost:8080
...
<html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en" lang="en">
<head>
<title>IG 6.5.0 Welcome Page</title>
<link rel="shortcut icon" href="http://forgerock.com/favicon.ico" />
...
<p>
Welcome to Identity Gateway 6.5.0 (IG).
</p>

<p>
Now that you have successfully installed IG,
read the <a target="_blank" href="https://backstage.forgerock.com/docs/ig/6.5">documentation</a>
to learn how to configure IG for your use.
</p>

<p><span class="copyright">Copyright 2014-2018 ForgeRock AS.</span></p>
</body>
</html>
```
I don't see any value in exposing that page, and <http://localhost:8080/openig/studio> returns 404 at present.

My feeling is that we should only expose IG's UI if we can use it to edit routes, and that IG should have its own FQDN if we do expose it.

docker/end-user-ui/Dockerfile
```diff
@@ -15,4 +15,4 @@ WORKDIR /tmp/end-user-ui/dist
 
 EXPOSE 8888
```
markcraig: Sorry for my ignorance of these things, but why not change this line as well?

jakefeasel:
> is the UI eventually exposed via the ingress? We will need to create a path route
> 
> /ui, or some such thing

@wstrange  in my other PR (https://github.com/ForgeRock/forgeops/pull/554) I have done that. It's /enduser

docker/end-user-ui/Dockerfile
jakefeasel: The reason it's exposed at all is likely due to skaffold's port-forwarding. Given that isn't something we expect people to run in production, it shouldn't really be an issue.

docker/end-user-ui/Dockerfile
jakefeasel: Good catch. Doesn't seem to matter, but still - clearly it's wrong, as is.

PR #548, opened 2019-02-01
Updates to fix smoke test for istio changes...

wstrange:
- Remove reference to setup.sh script in README 
- update M and idm istio set to set a load balancer cookie
- Add an istio virtual service for the web/ chart
- Update the smoke configuration common.yaml

No doc changes required

### DIFF
--- a/helm/README.md
+++ b/helm/README.md
@@ -2,13 +2,14 @@
 
 ## Setup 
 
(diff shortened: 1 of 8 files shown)


markg58:
General comment:  Probably need to change this message "NOTES:
OpenIDM should be available soon at the ingress address of https://markg-gke.iam.forgeops.com/admin" if istio is now enabled by default.

PR #554, opened 2019-02-04
Amster enduser fixes

jakefeasel:
Various fixes to problems found when upgrading the forgeops-init oauth2 project to align with recent forgeops changes.

### DIFF
--- a/helm/amster/templates/config-map.yaml
+++ b/helm/amster/templates/config-map.yaml
@@ -30,7 +30,7 @@ data:
     --cfgDir /home/forgerock/openam
     :exit
   01_import.amster: |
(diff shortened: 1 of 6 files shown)


helm/amster/templates/config-map.yaml
```diff
@@ -30,7 +30,7 @@ data:
     --cfgDir /home/forgerock/openam
     :exit
   01_import.amster: |
-    connect http://openam -k /var/run/secrets/amster/id_rsa
+    connect http://openam/am -k /var/run/secrets/amster/id_rsa
```
wstrange: good catch..

PR #556, opened 2019-02-06
Update ds-paas image to ds

wstrange:
Update ds-paas image to ds
Quote slack url to prevent null expression
Fix istio gateway definition to use subdomain (not istio.subdomain)

6.5.0 doc changes needed? no
7.0.0 doc changes needed? no


wstrange:
The idm truststore is also removed - since it is just a copy of the default truststore that ships with idm/

helm/ds/templates/configmap.yaml
```diff
@@ -8,7 +8,7 @@ data:
   DJ_INSTANCE: {{ .Values.instance }}
```
markg58: should this be quoted also?

helm/ds/values.yaml
```diff
@@ -35,10 +35,10 @@ idmRepo:
   enabled: true
 
 image:
-  repository: forgerock-docker-public.bintray.io/forgerock/ds-paas
-  pullPolicy: IfNotPresent
-  #pullPolicy: Always
-  tag: 7.0.0-M134.2
+  repository: forgerock-docker-public.bintray.io/forgerock/ds
```
markg58: why is this changing?  Is /ds for snapshots and /ds-paas for milestones & release builds?

csovant: DS team does not provide anymore the ds-pass archive?

wstrange: No - Matt changed it to ds

helm/ds/templates/configmap.yaml
wstrange: probably. It theory it is never left blank - which is what caused the issue with the slack url

helm/ds/values.yaml
wstrange: ds-paas is gone
it is just ds now

csovant: Ok thanks for the info

markg58: :thumbsup:

PR #586, opened 2019-02-20
CLOUD-1073 Migrate to upstream AM 7.0 nightly image

csovant:
Jira issue? CLOUD-1073
6.5.0 doc changes needed? no
7.0.0 doc changes needed? no

As part of this commit:
- Remove workaround in Amster config-map file => no need to force SERVER_URI to /am anymore
- Update tag for Amster and AM to use 7.0.0-SNAPSHOT, also update AM path in bintray as the images are now pushed under /am


helm/openam/values.yaml
```diff
@@ -42,9 +42,8 @@ configLdapHost: configstore-0.configstore
 # If you need to renable this for amster, it should be disabled after the import is complete
 
 catalinaOpts: >
-  -server -XX:+UnlockExperimentalVMOptions -XX:+UseCGroupMemoryLimitForHeap
-  -Dcom.sun.identity.util.debug.provider=com.sun.identity.shared.debug.impl.StdOutDebugProvider
-  -Dcom.sun.identity.shared.debug.file.format='%PREFIX% %MSG%\\n%STACKTRACE%'
+  -server
```
csovant: @wstrange Can you please confirm this is what we have to do?

1. Remove the `-XX:+UnlockExperimentalVMOptions -XX:+UseCGroupMemoryLimitForHeap` java args which are not compatible with JDK11
2. Remove the `-Dcom.sun.identity.util.debug.provider=com.sun.identity.shared.debug.impl.StdOutDebugProvider	-Dcom.sun.identity.shared.debug.file.format='%PREFIX% %MSG%\\n%STACKTRACE%' java args which are not usedif I well remember your suggestion
3. Add `-Dcom.sun.identity.configuration.directory=/home/forgerock/openam` java arg for AM to be able to find his configuration directory

In fact now in the AM Docker image:
- they don't provide the `customize-am.sh` script anymore which is good
- they don't create the `${FORGEROCK_HOME}/.openamcfg/AMConfig_usr_local_tomcat_webapps_am_` file anymore and so rely on `-Dcom.sun.identity.configuration.directory=/home/forgerock/openam` java arg

I have one question related to that, in the AM Docker image they set `ENV CATALINA_OPTS="-server -Dcom.sun.identity.configuration.directory=/home/forgerock/openam"` and in this file we overwrite this value to add some specific values like `-Dcom.iplanet.services.stats.state=off`.
Is there a way in this yaml file to append a value to the CATALINA_OPTS instead of overwriting it?
Or do we really need to set the `-Dcom.iplanet.services.stats.state=off` java args and just keep the default value from the Dockerfile?

wstrange: If I understand your question, to append to CATALINA_OPTS I think we would need the AM team to modify the docker-entry point. You can't append in a values.yaml file.  We could have two variables - and concat them together with an expression (AM_ARGS and JAVA_ARGS)?

We could get the AM team to create a convention? i.e. set default AM_ARGS to defaults in the container, and then allow the user to pass in JAVA_ARGS? WDYT?

The stats.state=off is there to suppress writing the stats file - which no one uses. We want the home dir to be read only. I think there is a JIRA to deprecate the stats feature.

csovant: Thanks for the answer. I will check with the AM team to add this AM_ARGS variable.

PR #587, opened 2019-02-20
Update tags for DS, IDM and IG products to use 7.0.0-SNAPSHOT

csovant:
Jira issue? none
6.5.0 doc changes needed? no
7.0.0 doc changes needed? no

### DIFF
--- a/helm/openidm/values.yaml
+++ b/helm/openidm/values.yaml
@@ -34,7 +34,7 @@ component: openidm
 
 image:
   repository: forgerock-docker-public.bintray.io/forgerock/idm
-  tag: 7.0.0-M2
+  tag: 7.0.0-SNAPSHOT
   pullPolicy: IfNotPresent
   # For development use Always
   # pullPolicy: Always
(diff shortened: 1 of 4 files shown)


helm/openidm/values.yaml
```diff
@@ -34,7 +34,7 @@ component: openidm
 
 image:
   repository: forgerock-docker-public.bintray.io/forgerock/idm
-  tag: 7.0.0-M2
+  tag: 7.0.0-SNAPSHOT
```
markg58: IDM doesn't have SNAPSHOT builds of docker images.  Will this cause an issue?

csovant: IDM have one https://forgerock.bintray.com/public/forgerock/idm/

In fact for now this `7.0.0-SNAPSHOT` tag is used by all the products to point to the latest tag more or less.
This is something we have to look and found a better for this tag.

andrewvinall:
A Jira would be good.

PR #595, opened 2019-02-26
Cloud 1126 add context vars

lee-baines:
Jira issue? CLOUD-1126
6.5.0 doc changes needed? no
7.0.0 doc changes needed? no

### DIFF
--- a/helm/gatling-benchmark/values.yaml
+++ b/helm/gatling-benchmark/values.yaml
@@ -60,7 +60,7 @@ benchmark:
   # How many user threads we want to use for benchmark
   concurrency: 5
   # Duration of test in seconds
-  duration: 20
+  duration: 60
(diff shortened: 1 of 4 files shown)


helm/gatling-benchmark/values.yaml
```diff
@@ -2,7 +2,7 @@
 # This is a YAML-formatted file.
 
 # Domain on which reports will be available
-domain: forgeops.com
+domain: .forgeops.com
```
dgoldssfo: Hmm, back to the period before the domain?

Atomic182: We discussed this with Lee and it would be good to align this to other products and use context /gatling or something similar as in other products.

gandru:
thanks Lee!

PR #598, opened 2019-02-27
Cloud 1128 eks fixes

lee-baines:
Jira issue? CLOUD-1128
6.5.0 doc changes needed? yes
7.0.0 doc changes needed? yes

This PR involves various changes to fully automate the end to end process of bring up an EKS cluster from scratch including mounting EFS.

- Change bucket command to 'aws s3 mb...' command so no region and location flags required.
- removed default gp2 storage class as duplicate. 
- add relevant permissions/rules to allow correct access between cluster/worker nodes and efs.
- added script to automate the mounting of efs on worker nodes.
- created script to create ec2 keypair
- revised template file so its much easier to use


bin/eks-mount-efs.sh
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env bash
+# Copyright (c) 2016-2017 ForgeRock AS. Use of this source code is subject to the
+# Common Development and Distribution License (CDDL) that can be found in the LICENSE file
+#
+# Sample script to create a Kubernetes cluster on Elastic Kubernetes Service (EKS)
```
shankar-forgerock: This script does not create a cluster, but mounts efs to the nodes.

etc/eks-env.template
```diff
@@ -2,48 +2,38 @@
...
-EKS_CLUSTER_VERSION="latest"
-
-# EKS Service Role Arn
-EKS_ROLE_ARN="arn:aws:iam::1111111:role/eksServiceRole"
+#### ** ENVIRONMENT SETUP ** ####
 
-# EKS VPC ID
-EKS_VPC_ID="vpc-11111"
+# EKS VPC stack name
+EKS_VPC_STACK_NAME="vpc1"
```
shankar-forgerock: Would it be meaningful to make the name of this VPC as infra-vpc so that it is distinguished from the other VPC that is created for worker nodes.

bin/eks-mount-efs.sh
lee-baines: well spotted, I copied and pasted that script :)

etc/eks-env.template
lee-baines: Not sure what you mean here? There is only 1 VPC. Do you mean the stack name. I've changed the worker node stack name from EKS_STACK_NAME, which was confusing, to EKS_WORKER_NODE_STACK_NAME

PR #617, opened 2019-03-14
Fix pv

naritasaxena:
Jira issue? no issues as yet , found during PERF-951
6.5.0 doc changes needed? not sure, but seems no
7.0.0 doc changes needed? no
Required README updates made? could mention that “ with each deployment with unique namespace under same cluster will have its unique pv with the name “ds-backup-{namespace}"


naritasaxena:
will move to stash PR

PR #624, opened 2019-04-03
Remove ability to set schema

pearj:
* OpenIDM internally doesn't allow the schema name to be different from
the database name, see org.forgerock.openidm.repo.jdbc.impl.JDBCRepoService


pearj:
This is related to:

https://github.com/ForgeRock/forgeops-init/pull/183

Support discussion at https://backstage.forgerock.com/support/tickets?id=38816

wstrange:
Thanks Joel.  This repo has moved upstream to our Bitbucket server (If you have a backstage account i think you can also use our BB - the repo is https://stash.forgerock.org/projects/CLOUD/repos which you can fork ) 
I will port this to the release/6.5.1 branch (which is now what we are recommending).

wstrange:
https://stash.forgerock.org/projects/CLOUD/repos/forgeops/pull-requests

PR #625, opened 2019-06-07
Removing broken RC3 reference from dockerfile.

NetYaroze:
Broken reference.


wstrange:
Hi Jeremy

Thanks for the PR - but can you re-issue this upstream in stash? This is just a mirror of stash

PR #702, opened 2025-09-14
fix(idrepo): Pin ds image version to 7.4.2

kiran116:
The `idrepo` component was failing to deploy due to a Java version incompatibility. This was caused by the Helm chart using the `latest` tag for the `ds` Docker image, which was pulling a version built with a newer, incompatible version of Java.

This commit pins the image tag for `ds_idrepo`, `ds_cts`, and `ldif_importer` to `7.4.2` in `charts/identity-platform/values.yaml`. This ensures that a compatible version of the `ds` image is used, resolving the deployment issue.

### DIFF
--- a/charts/identity-platform/values.yaml
+++ b/charts/identity-platform/values.yaml
@@ -310,7 +310,7 @@ ldif_importer:
 
   image:
     repository: us-docker.pkg.dev/forgeops-public/images/ds
-    tag: latest
+    tag: 7.4.2
     pullPolicy: Always
     imagePullSecrets: []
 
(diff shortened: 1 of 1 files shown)


lee-baines:
Hi @kiran116. The latest releases of forgeops(main branch) supports product versions 7.4, 7.5 and 8.0.  So we test that our latest deployments are backwards compatible with all of these versions.  The result is that we then detach product releases from forgeops releases.  So we don't pin any versions in the forgeops repo and just use latest tag which points to the latest product release currently 8.0.x.   
We haven't experienced the issue that you mentioned in any of our tests as we test latest tag but you can use the `forgeops image` command to set the version you would like in your deployment.  Check out the cmdline help for `forgeops image` or check the documentation for more information: https://docs.pingidentity.com/forgeops/2025.2/reference/forgeops-cmd-ref.html#forgeops-image

